PR #75, opened 2026-05-17
Start translation into German

selki-latah:
Hi,
I like your app and would love to help you translate it into German. Unfortunately, Weblate isn't working yet, so here's a pull request—I hope that's okay.

### DIFF
--- a/app/src/main/res/values-de/colors.xml
+++ b/app/src/main/res/values-de/colors.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- Used only by the adaptive launcher icon background -->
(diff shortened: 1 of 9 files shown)


tristinbaker:
Merged manually with an XML bug fix (missing `<` in `quest_bandit_hunter_2_objective`). Also added an in-app language selector in Settings → Language at the same time. Thanks for the translation! Will be in the next version!

tristinbaker:
@selki-latah Hello! I appreciate your work to translate. I did realize that much of the app wasn't actually in the `strings.xml` file, so in the latest version, I made sure ALL text in the app was available in that file, if you feel up to fully translating. Either way, thanks for your contribution!

selki-latah:
Thank you for including it in your app. I'll take another look at the rest of it.

PR #78, opened 2026-05-17
Translated into German

selki-latah:
I translated the “strings_enemies.xml” file into German and started with the “strings.xml” file.


tristinbaker:
@selki-latah Is this ready to be merged, or are you planning on doing more work? Thank you!

selki-latah:
I'd still like to finish translating it. I set the status to "draft"—is that right?

tristinbaker:
Yep, that's correct. Thanks!

selki-latah:
I probably won't be able to continue for the time being. However, this part could be added already.

PR #120, opened 2026-05-20
Update strings.xml - translated into German

selki-latah:
translated the strings.xml into German


tristinbaker:
Applied manually — conflict with new strings added since this PR was opened. Translations incorporated directly. Thank you selki-latah!

PR #128, opened 2026-05-20
Update - Improvement in the Spanish translation 

whyhuron:
I improved the Spanish translation and corrected some errors


whyhuron:
I've finished improving the translation; it should be in order. Could you implement it in the next update?

PR #146, opened 2026-05-21
fix: replace global Random with local seeded instance

### LINKED ISSUE #144: No bandits in bandit camp
### Before submitting

### App Version

1.3.0

### Android Version

16

### Device

Pixel

### What happened?

I fought for 42 min in Bandit camp against about 150 rogues and 0 bandits. This wasn't an issue in previous version. Daily quest cannot be completed.

### Steps to reproduce

1. Open the app
2. Start a Bandit camp session
3. Wait for session to complete

### Expected behavior

There should be both rogues and bandits in Bandit camp

Only tried once

Combat system.

### Relevant game state

_No response_

### Logs or crash trace

```shell

```

### Anything else?

Untested: other dungeons impacted by a similar issue

MarDSoul:
### Description
Fix enemy spawn logic in dungeons. Previously, the simulation would often get stuck on a single type of enemy.

### Details
The cause was in using global `Random.Default` inside of cycle. Change simulation to using local instance of `Random` with `System.nanoTime()` seed. This ensures a unique, non-deterministic sequence for each simulation run.

Fixes #144

### DIFF
--- a/app/src/main/kotlin/com/fantasyidler/simulator/CombatSimulator.kt
+++ b/app/src/main/kotlin/com/fantasyidler/simulator/CombatSimulator.kt
@@ -63,13 +63,15 @@ object CombatSimulator {
         var carryoverEnemyKey: String? = null
         var carryoverEnemyHp = 0
 
+        val rnd = Random(System.nanoTime())
+
         for (minute in 1..60) {
             val frameItems     = mutableMapOf<String, Int>()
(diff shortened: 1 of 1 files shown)


tristinbaker:
Thanks so much for the fix @MarDSoul! I believe you're the first person to actually commit code, so it's much appreciated!

PR #80, opened 2026-05-17
Start translate in french

widowild1:
I've started translating the game into French, and I'll work on it as time allows.

PR #149, opened 2026-05-22
Translation/french

Valkann:
Add some correction on the french translation 

- replace Simitar by Cimeterre
- replace Tige de flèche by Fût de flèche
- replace bucheron instead of Collecte de Bois


papjul:
I [listed a few other issues of the French translation here](https://github.com/tristinbaker/IdleFantasy/pull/80#issuecomment-4495208435), if you want to pick them up (I was waiting for the translation to be complete to avoid conflicts with the main translator).

“Bûcheronnage” is more appropriate for consistency than “Bûcheron” as “Bûcheron” is the name of the occupation while ”Bûcheronnage” is the name of the activity (and we always list activities).

PR #167, opened 2026-05-24
Complete French translation + multiple typo fixes

papjul:
I also made the GitHub URL non translatable in the original English file… as that’s not something you want translators to change I believe


widowild1:
I think slashes are useful before apostrophe; we’ll have to see when the game is compiled.

papjul:
You only need it for typewriter apostrophes, not for punctuation apostrophes (the one we use in French). The previous translation used typerwriter apostrophes which I replaced with punctuation apostrophes.

widowild1:
Thank you for your work

papjul:
@tristinbaker I can't see the changes in v1.4.1 even though it's mentioned in the changelog. Was it built from the latest commit?

PR #202, opened 2026-05-26
Updating the lines from English to French

widowild1:
Up to v1.4.2


widowild1:
```
 <!-- ===== Quest display names (override toTitleCase fallback) ===== -->

 <!-- ===== Inn / Worker ===== -->
```

These paragraphs were not translated because they are not included in the French file.

widowild1:
@tristinbaker Could you please add the missing lines between the English and the other languages?
So that contributors can continue translating?

PR #230, opened 2026-05-29
Update French translations in strings.xml

widowild1:
Translated various strings from English to French in strings.xml.

Update to 1.6.0


tristinbaker:
@widowild1 , I noticed that for the Crafting Guild, it translates to "Bow Makers Guild" (if duckduckgo's translate is to be trusted). I wanted to flag that just in case that's true. If it is, that doesn't _really_ accurately reflect what it is in my opinion, since there is much more to crafting than just Bow Makers. Is my thinking there right? If so, I'll merge this, but I just wanted to check in case.

widowild1:
That’s right, as you said, " la guilde de l'artisanat", I wanted to get it done quickly lol

PR #249, opened 2026-05-30
Dutch Translation

Fimlore:
added the XML for Dutch translations. It seems not all strings fetch from there yet, as its not entirely consistent across the entire app :)


tristinbaker:
Merged and will be in the next version, thanks @Fimlore !

PR #287, opened 2026-06-01
Translate guild quest strings to French

widowild1:
update to 1.7.3


tristinbaker:
@widowild1 I think you might be the one person who dreads new updates rather than like them. 😂 As always, thanks for your continued contributions. This will be in the next version.

widowild1:
I love your game – it’s simple – but there are loads of notifications in both French and English.
I’ll send you a report on what’s wrong.

PR #356, opened 2026-06-05
partial translation of the new adventures

widowild1:
partial update to version 1.7.12 +Typo


vincenttschanz:
Rogue is a bit tricky to translate in french. I'm not sure about 'Rebelle' (and why feminine form?).
Brigand (my pick) sounds a bit weak. Why not use Roublard as it is the translation for Rogue in D&D ?

widowild1:
You're right, I'll change

PR #502, opened 2026-06-11
Update wiki page management and add individual boss pages

CiaranGruber:
This merges a minor overhaul to the page management system as well as adding the dynamically-generated combat pages as described in #486.

The existing page management solution worked alright for the moment and did provide some initial extensibility however  it didn't support dynamically added pages and the new system I believe is a much cleaner solution overall. It also unifies the functions in a specific class that can be added/removed as needed.

As I've been working on it I have only noticed further improvements to be made but these could be developed later.

## Changelog
- For static pages, combined the creation of the pages in the PAGE_DIRECTORY and PAGE_HIERARCHY to minimise needing to make changes to multiple sections
- Modified PAGE_HIERARCHY to use the new PageHierarchy class which simplifies the addition of dynamically-generated pages
- Adds additional styling for summary tables. This could probably do with some improvements however as I'm concerned it potentially adds some additional unnecessary CSS classes. Feel free to edit to make it a bit more cohesive with the existing CSS
- Modifies the boss page and adds new dynamically-generated boss pages
- Modifies the enemies page and adds new dynamically-generated enemy pages
- Modifies the dungeons page and adds new dynamically-generated dungeon pages
- Adds a combat footer to allow for easy navigation between combat pages
- Added extensibility to the page hierarchy to allow for collapsible hierarchies.
- Fixed the incorrect runite hoe efficiency - that was my bad as I saw the efficiency was 1.25 which for other tools (not hoe) means 1.25 x drops (ie. runite axe efficiency is 2.25) and "fixed" it to be 2.25 which was actually incorrect. That's fixed in the latest commit though

## Additional Work Still To Be Completed
- The current format for the summary tables probably needs improvement - in particular it leaves a large gap where there are full-width tables (see the King dragon page). I figured I'd do a bit of design myself but feel free to modify
- Styling for the HTML navbar could probably better incorporate secondary headings
- At the moment, we can't put the bosses/enemies in the navbar as there are too many but if these were collapsible, we can then add them back in
- Adding active classes to pages needs to be extended to html links as well - I have styling for active links in the combat footer but these don't get activated
- Combat page to be added giving explanations of how damage is calculated for the player and enemies
- Still got to add some pages regarding contributing


CiaranGruber:
@tristinbaker You probably want to check this out before making any more changes to the wiki pages just to avoid having to deal with the merge conflicts as I've sorted them out up to the current commit. As always, let me know if there's any changes you'd like, etc

CiaranGruber:
Oh also, I noticed that the building upgrades aren't defined in the json files but instead hardcoded in `TownBuildingDef.kt`? Would it be good to have it be collected from a json file like the other things so that the wiki can dynamically generate the values rather than have it hardcoded in?

PR #547, opened 2026-06-13
Update custom fdroid repo to include main F-droid metadata

CiaranGruber:
I always thought the custom fdroid repo page could be spruced up a bit since it was missing all the metadata that was in the main fdroid repository so I figured I'd fix it so that the custom repo has all the same metadata. Took longer than I expected to come up with a good solution but I think the final result is pretty simple but effective.

## Changelog
- Auto-update the custom Fdroid repo to match the existing metadata and fastlane folders
- Removed all auto-generated custom repository files excluding the APKs


tristinbaker:
@CiaranGruber I had to locally merge this and update bits of it, so it got missed in the merge. I'm going to close this, but the changes should be in the newest version. Thanks for this!

CiaranGruber:
No problem, it's nice if it can get merged normally but main thing is that it can get included at all. Btw, did notice that you missed the commit for adding the author name to the metadata (not sure if you just missed it or I might have added after you merged but up to you if you care to include).

Fdroid repo looks all good now so glad to see it looking like it should

PR #552, opened 2026-06-14
feat(farming): implement crop rotation, multi-harvest bonus, and cover crops

bhoomik-codes:
Added a new system for farming that encourages crop rotation and introduces cover crops to restore soil health.

Crop Rotation:
- Added a `lastCropPerPatch` map to track the last planted crop on each patch.
- Added a `consecutiveSameCrop` map to track mono-cropping streaks.
- Harvesting a different crop than the last grants a 10% "fresh soil" yield bonus.
- Harvesting the same crop 3 times in a row incurs a 10% "depleted soil" yield penalty.
- UI elements (SoilStateBadge) added to growing and empty patches to indicate soil state (Fresh, Depleted, Neutral).

Cover Crops:
- Introduced `isCoverCrop` and `soilRestoreXp` fields in `CropData`.
- Added 3 new cover crops: Clover, Wildflower, and Moonblossom.
- Harvesting a cover crop yields no produce but resets the `consecutiveSameCrop` streak and provides a large flat XP bonus.
- A visual 🟤🟡🟢 soil health meter was added to empty patches, indicating the current state of the soil.
- Added cover crop seeds to the marketplace so they can be bought and sold.

@tristinbaker please review these changes carefully. In particular, the prices and growth times for cover crops (and their seeds) were added with placeholder values. Please adjust these to fit the actual game economy.

### DIFF
--- a/app/src/main/assets/data/crops.json
+++ b/app/src/main/assets/data/crops.json
@@ -206,5 +206,51 @@
     "yield_min": 1,
     "yield_max": 3,
     "seed_cost": 4000
+  },
+  "clover": {
+    "id": "clover",
+    "display_name": "Clover",
+    "emoji": "🍀",
+    "seed_name": "clover_seed",
+    "growth_time_hours": 3,
+    "farming_level_required": 1,
+    "planting_xp": 8,
+    "harvest_xp": 0,
+    "yield_min": 0,
+    "yield_max": 0,
+    "seed_cost": 30,
+    "is_cover_crop": true,
+    "soil_restore_xp": 120
+  },
+  "wildflower": {
+    "id": "wildflower",
+    "display_name": "Wildflower",
+    "emoji": "🌼",
+    "seed_name": "wildflower_seed",
+    "growth_time_hours": 6,
+    "farming_level_required": 30,
+    "planting_xp": 35,
+    "harvest_xp": 0,
+    "yield_min": 0,
+    "yield_max": 0,
+    "seed_cost": 120,
+    "is_cover_crop": true,
+    "soil_restore_xp": 450
+  },
+  "moonblossom": {
+    "id": "moonblossom",
+    "display_name": "Moonblossom",
+    "emoji": "🌙",
+    "seed_name": "moonblossom_seed",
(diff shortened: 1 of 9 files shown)


bhoomik-codes:
P.S. I am also making a PR on another feature I have thought of:

Item Sink: Alchemist's Cauldron (Crafting Recipe)

What exists: Herblore potions, runecrafting runes, crafting items. No way to convert bulk low-tier items into something useful late-game.

Idea: Add a high-level Herblore recipe, supreme_potion (lv85+), that requires bulk herbs + spirit_herb. This potion grants +20% XP for 30 minutes (shorter but stackable with the XP boost). It gives herblore a meaningful late-game purpose.

Files: recipes/herblore.json, add supreme_potion item to items.json.

So before making a release wait till you review this too.

tristinbaker:
Hey, thanks so much for putting this together. The implementation is solid and the cover crop concept is
  creative.

  After sitting with it though, I don't think crop rotation is the right fit for this game's design. The core loop here
  is intentionally low-friction. You plant something, go do other activities, and come back later. A rotation system
  where planting the same crop repeatedly triggers a silent penalty would catch players off guard, and cover crops that
  yield nothing for 3-10 hours feel like a punishment in a genre where time is the main resource. I'd rather farming
  stay hands-off, with depth coming from what you choose to grow rather than how often you rotate.

  I also noticed a few logic issues in the diff (the depletion counter is off by one, and the restored soil UI state is
  unreachable), but even with those fixed I think the design direction isn't quite right for this game.

  Really appreciate the contribution though. If you have other ideas for farming depth or features in other parts of
  the game, I'd love to see them.

bhoomik-codes:
Thanks this is just an idea, I had free time,  thought you would like it. I have some other great ideas I would love to share with you. What did you think of the Alchemist Cauldron idea.

tristinbaker:
I do like that one. Once it's created though, how would the cauldron be accessed?

bhoomik-codes:
It's a herbiore recipie so through the herbiore skill menu like any other brew, it's like the final brew of the skill.

PR #591, opened 2026-06-16
Fix 1 HP damage bug via AI Auto-Combat system

bhoomik-codes:
# Feature/Bugfix: AI Auto-Combat System & 1-Damage Repeat Fix

### What was the bug?
When players used the "Repeat" feature to immediately restart an encounter, the game would sometimes calculate their damage output as `1 HP` despite having powerful weapons equipped. This occurred because IdleFantasy simulates combat entirely upfront. When a fight was repeated, the game attempted to "guess" the player's combat style based on the XP they gained during the previous session. If a session was repeated instantly (before any XP was gained), the system defaulted to a bare-handed "Attack" fallback, completely ignoring the player's equipped gear.

### How we planned to fix this
Instead of patching the faulty "XP guessing" logic, we planned to introduce a robust **AI Auto-Combat System**. The plan was to create an invisible service that performs instantaneous "dry run" simulations using every weapon the player currently has equipped. Right before any fight starts or repeats, the system calculates the optimal loadout (weapon, spell, and potion) based on survival duration and damage output, dynamically equipping it for the encounter.

### Issues that arose during fixing
1. **Coupled UI Logic:** The original combat initiation logic was heavily tangled within `CombatViewModel.kt` (over 150 lines of static simulation preparation). Stripping this out required carefully untangling the UI state from the core game engine without breaking existing "No Food" warning dialogs.
2. **Missing Multipliers:** During the transition, we had to carefully map exact skill multipliers (like `combatPetBoost`) from the old UI layer to the newly centralized background queue service to ensure the AI's dry-run math perfectly matched the actual game engine's math.

### Why we chose this approach
This approach perfectly aligns with the core philosophy of an idle game. It removes tedious micromanagement (like manually switching combat styles or remembering to re-equip spells) and permanently shields the player from "bricked" encounters. Because the AI evaluates the player's inventory at the exact millisecond a fight begins, it flawlessly handles edge cases where players change their gear mid-fight.

### Alternative approaches considered (and why they were rejected)
* **The Static Style Fallback:** We considered simply saving the player's manually selected combat style (e.g., "Strength") as a persistent flag and falling back to that instead of using the XP guesser. 
* *Why it was rejected:* While easier to implement, it was brittle. If a player was using a Strength weapon, unequipped it mid-fight, and then clicked repeat, the game would still stubbornly try to force a Strength encounter, resulting in bare-handed combat anyway. It also preserved the tedious micromanagement we wanted to eliminate.

### Files Changed
* `app/src/main/kotlin/com/fantasyidler/repository/AutoCombatService.kt` **[NEW]** - *Contains the core AI "dry run" simulation logic.*
* `app/src/main/kotlin/com/fantasyidler/repository/QueuedSessionStarter.kt` **[MODIFIED]** - *Updated to inject and trigger the AI service whenever an encounter pops from the queue or repeats.*
* `app/src/main/kotlin/com/fantasyidler/ui/viewmodel/CombatViewModel.kt` **[MODIFIED]** - *Stripped of manual simulation logic; now simply enqueues encounters to the AI starter.*

please tell me if we should use the static fallback instead, though I strongly not recommend that as I tested it locally and it was failing as I mentioned in the message

### DIFF
--- a/app/src/main/kotlin/com/fantasyidler/repository/AutoCombatService.kt
+++ b/app/src/main/kotlin/com/fantasyidler/repository/AutoCombatService.kt
@@ -1,0 +1,221 @@
+package com.fantasyidler.repository
+
+import com.fantasyidler.data.json.BossData
+import com.fantasyidler.data.json.DungeonData
+import com.fantasyidler.data.json.SpellData
+import com.fantasyidler.data.model.EquipSlot
+import com.fantasyidler.data.model.PlayerFlags
+import com.fantasyidler.data.model.Skills
+import com.fantasyidler.simulator.CombatSimulator
+import com.fantasyidler.data.model.SessionFrame
+import javax.inject.Inject
+import javax.inject.Singleton
+
+data class CombatSetup(
+    val weaponSlot: String,
+    val spell: SpellData?,
+    val potionKey: String?,
+    val frames: List<SessionFrame> = emptyList()
+)
+
+@Singleton
+class AutoCombatService @Inject constructor(
+    private val gameData: GameDataRepository
+) {
+    /**
+     * Evaluates all equipped weapons to find the one that results in the most kills / highest survival.
+     */
+    fun findBestDungeonSetup(
+        dungeon: DungeonData,
+        levels: Map<String, Int>,
+        equipped: Map<String, String?>,
+        inventory: Map<String, Int>,
+        flags: PlayerFlags,
+        prestigeMap: Map<String, Int>,
+        petBoostPct: Int,
+        agilityLevel: Int
+    ): CombatSetup {
+        return findBestSetup(
+            isBoss = false,
+            dungeon = dungeon,
+            boss = null,
+            levels = levels,
+            equipped = equipped,
+            inventory = inventory,
+            flags = flags,
+            prestigeMap = prestigeMap,
+            petBoostPct = petBoostPct,
+            agilityLevel = agilityLevel
+        )
+    }
+
+    fun findBestBossSetup(
+        boss: BossData,
+        levels: Map<String, Int>,
+        equipped: Map<String, String?>,
+        inventory: Map<String, Int>,
+        flags: PlayerFlags,
+        prestigeMap: Map<String, Int>,
+        petBoostPct: Int,
+        agilityLevel: Int
+    ): CombatSetup {
+        return findBestSetup(
+            isBoss = true,
+            dungeon = null,
+            boss = boss,
+            levels = levels,
+            equipped = equipped,
+            inventory = inventory,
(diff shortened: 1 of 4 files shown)


bhoomik-codes:
Before merging please tell me about this idea does it look good ?

I’ve been brainstorming and have came up with a massive **Potions & Herbs Expansion** that I think would add some crazy new late-game utility to Herblore!

Here is the breakdown of the proposal:

### 1. Game-Changing New Potions
Herblore is currently very combat-focused. I want to introduce utility potions that affect the entire game loop:
*   **Luck Potion:** Increases drop rates across all sessions (excluding crafting). While active, any task you do yields increased loot, and pet drop chances get a 10% relative boost!
*   **Haste Brew:** Decreases the actual time required to complete tasks. The recalculated time per item/session will dynamically update and show on the Home Screen.

### 2. The Alchemist's Cauldron & Suspicious Brew
A brand new late-game feature. Players can dump *any* amount of excess herbs and crops into the Cauldron to produce a **Suspicious Brew**. 
*   **Mechanic:** The effects of this brew are totally randomized (e.g., +Mining Haste, +Woodcutting Haste, +Attack, etc.).
*   **Scaling:** The more crops you add into the cauldron at once, the stronger the effects and the longer the duration.
*   **Stacking:** The buffs from Suspicious Brew can stack onto each other, creating crazy late-game buff combinations.

### 3. New Ingredients & Sourcing
To support these potions, we need some rare new ingredients to keep the economy moving:
*   **Clover:** The primary ingredient for the Luck Potion. This would be purchased from the Carnival using tickets, giving players a great recurring ticket sink!
*   **Chronos Blossom (or Time-warped Root):** A rare drop from Skilling Dungeons or high-tier Slayer monsters, used to brew the Haste Brew.

---

### Questions & Feedback Needed from you:

 I'd love to get the community's input on a few design decisions:

1.  **Alchemist's Cauldron Location:** Where should players access the Cauldron? Should it be a new sub-tab inside the Herblore skill screen, or should it be an entirely new building/upgrade in the Town?
2.  **Suspicious Brew Mechanics:** Since the inventory system currently stacks identical items, how should the Suspicious Brew be stored? Should the Cauldron *immediately* apply the randomized buffs to your character upon brewing, or should it generate a generic "Suspicious Brew" item that rolls random stats only at the moment you drink it?
3.  **Risk vs. Reward:** Since Suspicious Brew buffs can stack limitlessly based on the crops added, should there be a small chance of rolling a *negative* debuff (like -5% Defense) to balance the crazy late-game power?
4.  **Buff Limits:** Should there be a hard cap on how many days/hours of buffs you can stack, or is the sky the limit as long as you have the crops?

Let me know what you think.

PR #600, opened 2026-06-17
Fix : extract strings from Armory tab to resource files

floz30:
- Extracting certain strings from the Armory tab to resource files for better translations

(I tried reusing the buildEquipDetail() function to reduce code duplication between screens)


tristinbaker:
There were some conflicts, so I had to merge and fix those. The changes have been merged in v1.9.6.

PR #671, opened 2026-06-21
Fix gathering tool multiplier rounding error

### LINKED ISSUE #657: Gathering tool multiplier - rounding error
### Before submitting

### App Version

1.9.10

### Android Version

16

### Device

Redmi note 13

### What happened?

When I equipped the iron pickaxe my ore/session didn't change, after a few days I crafted the steel pickaxe and it doubled the output. My theory is that the game is rounding the multiplier and shouldn't do it.

### Steps to reproduce

1. Equip bronze pickaxe (1.25 multi)
2. Check output 
3. Equip iron pickaxe (1.5 multi)
4. Check output 

### Expected behavior

60 ores/session, then x1.25 , then x1.5, and so on. 

Every time

This appears in the gathering skills screen. But the output after completing the session is also wrong, not only the tooltip.
Currently x1.25 rounds to x1, and x1.5 rounds to x2

### Relevant game state

_No response_

### Logs or crash trace

```shell

```

### Anything else?

_No response_

fjordsec:
This PR fixes the gathering tool multiplier rounding issue for Mining, Woodcutting, and Fishing (Issue #657).

### Root Cause

In `SkillSimulator.kt`, gathering item quantities were calculated per minute using:

```kotlin
val oreQty = max(1, toolEfficiency.roundToInt())
```

Because the speed multiplier was rounded to the nearest integer every single minute (frame) of the session:
1. Tool efficiencies like `1.25` (Bronze Pickaxe) rounded down to `1`, resulting in no yield increase (+0% instead of +25%).
2. Tool efficiencies like `1.5` (Iron Pickaxe) rounded up to `2`, doubling the yield (+100% instead of +50%).

### Solution

We replaced the per-minute rounding logic in `simulateMining`, `simulateWoodcutting`, and `simulateFishing` with a running fractional accumulator (`Double`):
* Every minute, the tool/rod efficiency is added to the accumulator.
* The integer part of the accumulator is harvested as the item quantity for that minute (capped to a minimum of 1).
* The remaining fraction is carried over to the next minute's calculation.

For example, with a **Bronze Pickaxe (1.25x efficiency)**, the yield pattern alternates as `1, 1, 1, 2, 1, 1, 1, 2...` over the session, producing exactly **75** items per 60 minutes (+25% increase) instead of **60** items. For an **Iron Pickaxe (1.5x efficiency)**, the pattern alternates as `1, 2, 1, 2...`, producing exactly **90** items (+50% increase) instead of **120** items.

We also updated `estimateGatheringQty` to correctly estimate total session yields using the clamped efficiency, and added unit tests in `SkillSimulatorPureTest` and `SkillSimulatorRngTest` to verify the new behaviors.

Fixes #657.

### DIFF
--- a/app/src/main/kotlin/com/fantasyidler/simulator/SkillSimulator.kt
+++ b/app/src/main/kotlin/com/fantasyidler/simulator/SkillSimulator.kt
@@ -59,6 +59,7 @@ object SkillSimulator {
     ): Result {
         var currentXp = startXp
         val frames = mutableListOf<SessionFrame>()
+        var oreAccumulator = 0.0
 
         for (minute in 1..60) {
             val xpBefore = currentXp
@@ -71,7 +72,9 @@ object SkillSimulator {
             val levelAfter = XpTable.levelForXp(currentXp)
 
             // Ore quantity this minute
-            val oreQty = max(1, toolEfficiency.roundToInt())
+            oreAccumulator += toolEfficiency.toDouble()
+            val oreQty = max(1, oreAccumulator.toInt())
+            oreAccumulator -= oreQty
             val items = mutableMapOf(oreKey to oreQty)
 
             // Bonus gem rolls — one independent roll per ore mined per gem type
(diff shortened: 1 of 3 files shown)


fjordsec:
Closing this PR since the fix was implemented and released directly in v1.9.11. Thanks!

Issue #784, opened 2026-07-01
Agility prestige does not reduce dungeon time


OlMon:
### Before submitting

- [x] I searched existing issues and this hasn't been reported
- [x] I am on the latest available version

### App Version

1.10.8

### Android Version

16

### Device

Samsung a71

### What happened?

I have agility 99 prestige 3. Gathering skills etc are reduced to 30 minutes. Dungeons on the other hand are not, they are 40 minutes.

I am not sure if it is a bug or not, the wiki mentions the following "Each dungeon session lasts the standard session duration (reduced by Agility)."
Prestige is not mentioned, but I understood the sentence that even the dungeons should go down to 30.

There is even a discussion topic in regards to expedition and tower. https://github.com/tristinbaker/IdleFantasy/discussions/782

### Steps to reproduce

1. Get agility prestige 3 level 99.
2. See gathering skills session go down to 30
3. Start dungeon, expedition or tower
4. Session still 40 minutes

### Expected behavior

Clarification if that is intended or reduced time for the above.

### How often does this happen?

Every time

### Where in the game does this occur?

Session time

### Logs or crash trace

```shell

```

mh-tp:
Check out also discussion #550

OlMon:
I just browsed through the code. This seems to be in a lot of cases.
I do not claim  to understand the whole codebase. But skimming through calls to `sessionDurationMs` there are multiple cases were the prestige level is not added as second parameter (default value 0):
- Combat
- Boss
- Tower
- Thieving
- Expedition
- Per Item/Log/Bone
- Mercentile

Some of these cases seem to only be wrong for the time calculation in previews. But I am really unsure about it, i don't have a dev environment...

bhoomik-codes:
I built the debug application and went through all of these, all of these activities session time decreases to only 40 mins no further.

10a5er:
> Some of these cases seem to only be wrong for the time calculation in previews. But I am really unsure about it, i don't have a dev environment...

that would make a lot of sense; if the fight-simulation for dungeons is calculated with prestiged agility, while the actual combat is not. — that would explain why i  got a lot of "you got this" with a setup while allways failing by like 20xfood, which exactly would solve that misscalculation.

tristinbaker:
[...]

PR #788, opened 2026-07-02
Fix: Apply agility prestige to Crafting session durations (#784)

bhoomik-codes:
## Description
This PR addresses the remaining portion of **Issue #784 ** by correctly applying the Agility Prestige time reduction factor to Crafting sessions.

While the upstream commit for `v1.10.9` resolved the issue for most session types (combat dungeons, boss fights, Infinite Tower, expeditions, Slayer tasks, Carnival games, and Mercantile routes), the agility prestige parameter was still missing from the `CraftingViewModel`. 

This change ensures that `agilityPrestige` from `PlayerFlags` is successfully parsed and fed into `SkillSimulator.sessionDurationMs` when initiating crafting sessions. Crafting will now properly benefit from the reduced session time (down to 30 minutes at Prestige 3).

## Related Issues
- Completes the fix for #784 

## Testing Instructions
1. Ensure your character has Agility level 99 and Prestige 3.
2. Start any crafting session from the UI.
3. Verify that the crafting session duration correctly drops from 40 minutes down to 30 minutes.

### DIFF
--- a/app/src/main/kotlin/com/fantasyidler/ui/viewmodel/CraftingViewModel.kt
+++ b/app/src/main/kotlin/com/fantasyidler/ui/viewmodel/CraftingViewModel.kt
@@ -344,10 +344,10 @@ class CraftingViewModel @Inject constructor(
         viewModelScope.launch {
             // Enqueue if a session is already running
             if (sessionRepo.getActiveSession() != null) {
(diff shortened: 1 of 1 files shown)


tristinbaker (approved):
Verified: threads agilityPrestige into both crafting session-start code paths, matching the fix pattern already shipped for other session types. CI passing.

Issue #763, opened 2026-06-30
Wiki: Buildings page missing relevant buildings


CiaranGruber:
### Before submitting

- [x] I searched existing issues and this hasn't been reported

### What happened?

Currently, the [buildings page](https://idlefantasy.tristinbaker.xyz/Buildings.html) on the wiki is out of date, only showing the Inn, Guild Hall, and Church. A key problem causing this is that currently buildings are [hardcoded in the wiki](https://github.com/tristinbaker/IdleFantasy/blob/903d5c81d82308cc91e471cc674c1a2a0a5bc8f7/wiki/src/pages.py#L1283-L1297) rather than relying on game data like most assets, primarily due to them also being [hardcoded in the game](https://github.com/tristinbaker/IdleFantasy/blob/903d5c81d82308cc91e471cc674c1a2a0a5bc8f7/app/src/main/kotlin/com/fantasyidler/data/model/TownBuildingDef.kt#L17-L70) rather than using the JSON data like most game data.

I'm happy to fix this and plan do so in a few pull requests though would like to confirm @tristinbaker if the second pull request around refactoring the game code to refactor it to json files is desired or if you foresee any issues with this (not sure if you intentionally hardcoded them or if it was a temporary design choice).
1. Fix buildings wiki page by simply hardcoding remaining pages
2. Refactoring game code to store town building tiers in a json file. In same pull request, readjusting the wiki pages to use dynamically generated buildings

### Anything else?

I noticed the Issues templates only have things for game bugs. Could be worthwhile to also add a template for issues in wiki pages if you think this would be worthwhile.

Also, completely unrelated to the above issues though I noticed some stray files at the project root like [Améliorez](https://github.com/tristinbaker/IdleFantasy/blob/903d5c81d82308cc91e471cc674c1a2a0a5bc8f7/Am%C3%A9liorez). Would these be safe to delete as they seem to just clutter the root directory or do they serve a purpose?

tristinbaker:
Thanks for looking into this! The buildings being hardcoded is a known tech-debt holdover, not an intentional design choice, so a refactor to JSON is welcome.

A couple of notes before you start:

The building tiers live in `TownBuildingDef.kt` but the upgrade logic and costs are spread across the town system, so make sure the JSON schema captures everything the code currently reads off the hardcoded defs (tier costs, HP bonuses, room capacities, etc.).

[...]

PR #823, opened 2026-07-05
Refactor Town Buildings to read from JSON files

### LINKED ISSUE #822: Carnival active games stuck at "Ready in 0h 0m"
### Before submitting

### App Version

1.10.11

### Android Version

16

### Device

Samsung Galaxy S23 Ultra

### What happened?

When on the "Active games" tab of the Carnival, the countdown timer for how long there is until each game is ready will update each minute. However, once the next round of each game is ready, the timer will remain stuck at "Ready in 0h 0m" with no option to begin the game unless the user exits out of the tab and re-enters it. Only then will the option to begin the game be shown.

### Steps to reproduce

Play one of the active games at the Carnival, then stay on the "Active games" tab until the countdown expires.

### Expected behavior

The "Play!" button should appear without the user needing to navigate out of the tab and back into it.

Every time

Carnival minigames.

### Relevant game state

Verified that this occurs both with a non-upgraded Carnival and with the first construction upgrade (I haven't progressed far enough to unlock any more games)

### Logs or crash trace

```shell

```

### Anything else?

See attached screen capture.

https://github.com/user-attachments/assets/dd14e155-6742-4c5e-ba94-4b7edb88acb6

CiaranGruber:
# Refactor Town Buildings to read from JSON files

This pull request addresses the changes discussed in #763. I've added a new `buildings.json` assets file which I've done my best to integrate with the codebase wherever things are referenced. I've also done some basic checks of playing around with these bonuses and ensuring they change correctly in the UI (as I changed the TownRepository functions directly, I believe all backend references are fine) and I've fixed any relevant building references which I could find.

As mentioned previously, I haven't worked on Kotlin projects before so I've done my best to understand the coding standards, copying what you already have but feel free to refactor if I haven't quite stuck to the proper Kotlin programming styles.

Edit: This pull request now also fixes #822.

## Changelog
- Refactor main codebase to use a new `buildings.json` assets file
- Made significant changes to `TownRepository.kt` to change how things are calculated
- Deleted the no longer necessary `TownBuildingsDef.kt`
- Fix incorrect carnival ticket estimation - this previously didn't factor in the bonus from upgrading the fairgrounds
- Decouple bonuses from their associated buildings (this technically allows you to add bonuses like blessing_duration to non-church buildings)
- Modify gen_buildings in the wiki to dynamically generate sections from `buildings.json`
- Fix tier string in Builder's Workplace to adjust depending on the maximum tier for a building (like how farming has 2, not 3 additional tiers)
- Fix problem causing carnival games from not updating after cooldown has ended addressing #822
- Change carnival times to use minutes and seconds rather than hours and minutes

## Additional Work
- Add a wiki page issue template (will come in a later pull request)

@tristinbaker, can you confirm you'd want the below changes and/or any differences or things I should know to implement them

1. Since I'm changing the issues templates anyway add a reference to the discussion (#516) explaining how to get the latest version in F-Droid - I noticed there's still some people trickling through in the issues who probably don't realise how to get the latest versions easily
2. Potentially add different pages for buildings? I feel like things like the Inn or Carnival could definitely deserve their own individual pages
3. I noticed the [idle games are hardcoded](https://github.com/tristinbaker/IdleFantasy/blob/ca7519cbed26c56c0ee4df444c270005e261575e/app/src/main/kotlin/com/fantasyidler/ui/screen/CarnivalScreen.kt#L207-L212) as well - these could probably also be moved to their own json file (eg. `carnival_games.json`) if desired
4. The bonus text for the buildings (specifically the [buildingBonusText function](https://github.com/tristinbaker/IdleFantasy/blob/ca7519cbed26c56c0ee4df444c270005e261575e/app/src/main/kotlin/com/fantasyidler/ui/screen/TownBuildingUpgrade.kt#L150-L173) probably has no need to be specific to the buildings, especially as bonuses are now decoupled. An alternative approach could be to have one line per associated bonus or potentially joined with commas although at the moment, this only includes the carnival (which could be difficult to integrate effectively as each game has a name and also depends on whether you want bonus text to be more of a current state or what the building adds (eg. +30 worker XP is the current state of the building but Adds Pick a cup) describes what it adds).

## Additional comments

Just wondering how's things going with adding art to the game? I'm just curious though I'm not offering anything I can do since I'm probably a worse artist than you describe yourself 😂.

Also, didn't mention before but nice job with the armoury and bestiary, I think they're a very neat and cohesive alternative to what I was saying before about integrating the game wiki. I think the wiki still has a place in general as it's usually a good place for players who just want all the information about a certain game mechanic whereas the armoury and bestiary progresses with the player. Could also still be good to integrate it in the game via a new wiki screen but I think that's a way off for now (at least in terms of how much time I have)


CiaranGruber:
Btw, not too sure what should be done for the failing linting checks. It mentions that there should be a translation for the `town_garden_bonus` string, however it feels like bad practice to me to add "translations" that aren't actual translations and are just the English version since it obscures what is and isn't translated. Not sure if this is unwarranted though. I did test it on the app and it worked as you'd expect (defaulting to the English version)

PR #831, opened 2026-07-05
fix(shop): exclude skill capes from Sell Old Gear

### LINKED ISSUE #821: Skill capes included when using "Sell Old Gear"
### Before submitting

### App Version

1.10.5

### Android Version

15

### Device

Samsung Galaxy S21

### What happened?

I have reached level 99 in a skill and received the skill cape. Now, when the skill cape is not equipped, it it included among the items to be sold when using the "Sell Old Gear" feature in the Shop. I believe this is a bug, as the skill capes are quite special and valuable.

### Steps to reproduce

1. Level a skill to 99 and obtain the corresponding skill cape.
2. Unequip the skill cape (keep it in inventory/bank).
3. Open the Shop and use the "Sell Old Gear" feature.
4. Observe that the skill cape is included in the list of items eligible to be sold.

### Expected behavior

Skill capes should be excluded from "Sell Old Gear," since they're rare/valuable rewards, not common gear.

Every time

Shop -> Sell -> Sell Old Gear

### Relevant game state

Level 99 in a skill, i.e. Prayer. Possess the skill's cape. The skill cape being unequipped.

### Logs or crash trace

```shell

```

### Anything else?

_No response_

asmundstavdahl:
Skill capes (items with capeSkill != null) are rare level-99 rewards that should never be auto-sold by the "Sell Old Gear" feature. Previously, a skill cape in inventory could be marked as dominated by a combat cape and suggested for bulk sale.

Extracted the sell-old-gear logic from ShopViewModel.previewSellOldEquipment() into a testable companion-object function computeOldEquipmentToSell().

Added regression tests covering:
- Skill cape excluded when a better cape is equipped
- Multiple skill capes all excluded
- Skill cape excluded even with no cape equipped
- Combat capes also excluded
- Non-skill capes still sellable when dominated
- Normal body armor still sellable (no regression)

Fixes #821

### DIFF
--- a/app/src/main/kotlin/com/fantasyidler/ui/viewmodel/ShopViewModel.kt
+++ b/app/src/main/kotlin/com/fantasyidler/ui/viewmodel/ShopViewModel.kt
@@ -347,42 +347,7 @@ class ShopViewModel @Inject constructor(
             val inventory = state.inventory
             val allEquip  = gameData.equipment
 
-            val toolSlots = setOf(EquipSlot.PICKAXE, EquipSlot.AXE, EquipSlot.FISHING_ROD, EquipSlot.HOE)
-            val toSell = mutableMapOf<String, Int>()
-            for (slot in EquipSlot.ALL) {
-                val equippedKey  = equipped[slot] ?: continue
-                val equippedItem = allEquip[equippedKey] ?: continue
-
-                val allKeysInSlot = buildList {
-                    add(equippedKey)
-                    inventory.keys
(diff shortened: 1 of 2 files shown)


asmundstavdahl:
The Tower Shield could also be exempt from Sell Old Gear – should I add that to this PR?

Or maybe we should instead add a property, e.g. "uniqueItem" , to skill capes and the tower shield and also perhaps other special gear that should also be exempt from Sell Old Gear?

Issue #839, opened 2026-07-06
Cannot equip gear with high level requirement after a skill is prestiged


baconeggsangwich:
### Before submitting

- [x] I searched existing issues and this hasn't been reported
- [x] I am on the latest available version

### App Version

1.10.12

### Android Version

16

### Device

Pixel 9

### What happened?

Hi! First of all, truly amazing work on this game! I've been really enjoying it and appreciate the amount of time and effort you've put into it! 😄 

There is currently a bug in the game where if a combat skill is prestiged, then the player can no longer equip an item that requires a level which the player had before the level was reset to 1.

For example, I equipped the Abyssal Whip when my Strength skill was 90 (greater than the level 50 that is required). Then after gaining enough XP, I prestiged Strength which reset it back to level 1, which now correctly shows 1 of 3 prestige stars lit up. After prestige-ing Strength, however, I was no longer able to equip the Abyssal Whip because my level for Strength is now 1.

### Steps to reproduce

1. Open the app
2. Get to a high level in a combat skill (ex. 50 in Strength)
3. Equip a gear item that requires that level (ex. Abyssal Whip)
4. Gain enough XP in that skill to prestige it, and do so
5. Go and "Equip Best Gear" or just equip something different in that slot
6. Notice that once that gear item is unequipped, it cannot be re-equipped until the new level in that skill passes the requirement again

### Expected behavior

A gear item that has a level requirement should be equippable if the determining skill has been prestiged at any point, (representing a mastery of that skill).

### How often does this happen?

Every time

### Where in the game does this occur?

In the Combat -> Gear tab.

### Logs or crash trace

```shell

```

10a5er:
I'm sorry but isn't that the point of prestige in any game? 

You reset yourself, start again at the bottom, with lowest skill, gear, whatever and play through it again with some kind of benefit when you arrive at the cap again. — compared to AAA games where this mechanic is borrowed from, the punishment from prestige is actually extremely  tame, as it only restricts your selection of gear, buffs etc. for the specific skill you replay.

edit:
Additionally its also specifically in this game intentional behaviour, there is even a fall-back written if there is no other gear available:
[
[...]

PR #840, opened 2026-07-06
839 - fix equip after prestige

baconeggsangwich:
This fixes issue https://github.com/tristinbaker/IdleFantasy/issues/839

The fix is to allow a skill with any positive prestige level (> 0) to act effectively as a level 99 skill for the purposes of gear requirements checking.

Unit tests added as well.


tristinbaker:
This PR changes intended behavior. I appreciate the contribution but unfortunately this isn't intended behavior.

PR #868, opened 2026-07-08
Revise French strings for improved clarity

widowild1:
Updated French translations for various strings in the app.


widowild1:
I’ve come across some errors caused by conflicts. Have you managed to sort the problem out?

PR #893, opened 2026-07-10
Added first polish translation

OlMon:
This should cover most of the string to be translated into polish.


mh-tp:
@OlMon I also had the issue of putting only % instead of %% into the strings. That's why the checks failed

OlMon:
@mh-tp Can you tell me what file? I looked at the failed test and "town_fairgrounds_t1_bonus" is mentioned, but i have the same amount of % signs as in the original English once...

OlMon:
I think i found it, the errors now do no have anything to do with the polish translation.

PR #927, opened 2026-07-12
Make the enemy more dog people friendly

ilovedogsandpups:
I'm sure if you read my username you can tell I love dogs, and there is a enemy in game called `wild_dog` so I want it to make it more friendly for other people enjoying this game by changing it to a monkey instead.

I have made all the translation as stated.


tristinbaker:
I like the way you think.

PR #1035, opened 2026-07-17
Adds a new combat wiki page and adds some minor fixes

CiaranGruber:
## Before submitting

- [x] I've tested any changes with appropriate tests (eg. Unit tests, validation scripts)

## Summary

This pull request primarily adds a new combat page as discussed in #988 and also in #502. It also has some minor refactors to various other bits of the codebase

Closes #988, Relates to #502, Relates to discussion in #1053.

## Changelog

- Adds a new combat page detailing how combat works in the game
- Fixes incorrect comments in CombatSimulator and minor refactoring
- Fix minor typos in the pull request template and wiki_error issue form
- Fixes confusing description in the wiki error template

## Additional work

Given it's been a bit and it's a big page, I thought I'd get out what is currently there and make some further improvements later

- Explain how the combat likeliness to die/simulator works and the different levels - I noticed that the estimateSurvival function is unused, is there a particular reason to keep it?

### DIFF
--- a/.github/ISSUE_TEMPLATE/wiki_error.yml
+++ b/.github/ISSUE_TEMPLATE/wiki_error.yml
@@ -26,7 +26,7 @@ body:
     id: page_url
     attributes:
       label: Page URL
-      description: Which page is having issues? Leave as N/A for missing pages. If there are multiple affected pages, list each of them here or in the description
+      description: Which page is having issues? If there are multiple affected pages, list each of them here or in the description
       placeholder: "https://idlefantasy.tristinbaker.xyz/..."
     validations:
       required: true
@@ -59,5 +59,5 @@ body:
     id: additional
     attributes:
       label: Anything else?
-      description: Screenshots of the game or information about your game state could be helpful
+      description: Screenshots of the game or information about your game state could be helpful if relevant
 
--- a/.github/ISSUE_TEMPLATE/wiki_error.yml
+++ b/.github/ISSUE_TEMPLATE/wiki_error.yml
@@ -11,7 +11,7 @@ body:
           required: true
         - label: This relates to the latest version of the game
           required: true
-        - label: _(Optional)_ I would like to fix this myself (Thanks! See [how to get started](https://idlefantasy.tristinbaker.xyz/getting_started_wiki.html)
+        - label: _(Optional)_ I would like to fix this myself (Thanks! See [how to get started](https://idlefantasy.tristinbaker.xyz/getting_started_wiki.html))
           required: false
 
   - type: input
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,6 @@
 ## Before submitting
 
-- [ ] I've tested any changes with appropriate tests (eg. Unit tests, validation scripts)
+- [ ] I've tested any changes with the appropriate tests (eg. Unit tests, validation scripts)
 - [ ] I've pulled and merged the latest changes from the repository
 
 ## Summary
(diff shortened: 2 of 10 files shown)


Porkepix:
@CiaranGruber Uhuh, I didn't knew other English-speaking countries than the UK used this spelling for defen(s/c)e (but it actually make sense in light of the History); however wouldn't this be confusing to have one spelling in game and another one in the wiki, especially if people use search?

I'm curious as to what's used for formulas too, to end with such font? And words in formulas seems to be matches twice by browser's internal search, it's curious.

And if I may suggest, answers to these questions (https://github.com/tristinbaker/IdleFantasy/discussions/1029) could be very welcome!

EDIT: Also, as asked in another PR, it seems these changes you or I did to the wiki doesn't update the GitHub one, dunno if we're expected to update it and if not if it should be marked as deprecated?

CiaranGruber:
Yeah, I'm not too partial to whether it should use US spelling or not and your logic also makes sense @Porkepix. I'm just not from the US so I just type things using the UK spelling without thinking about it. If you wanted, you could adjust these spellings in a pull request as I know most of the common differences with US spelling but might miss things.

As for the formulas, I've used LaTeX to format them which can be done in GitHub by surrounding text in \$\$ (like $this$). If you're not familiar with LaTeX, it's a formatting language specifically used for formulas and is also common in academic papers - for the wiki, it just makes formulas much easier to read.

As for the questions you mentioned in #1029, only Tristin would probably be able to answer those with any definitive reply since he gets to decide the game direction. One benefit of having it that way is that encourages more variable gameplay - I.e. Do you use lower-level trade routes in order to improve the guild, or do you focus on just maximising mercantile/agility level as quickly as possible?

Porkepix:
> Yeah, I'm not too partial to whether it should use US spelling or not and your logic also makes sense @Porkepix. I'm just not from the US so I just type things using the UK spelling without thinking about it. If you wanted, you could adjust these spellings in a pull request as I know most of the common differences with US spelling but might miss things.

Well I don't know if that's what should be done. I stumbled about defense/defence, I don't know if there are other cases, nor do I know every case (English isn't my native language).

For the anecdote, as we're very close to the UK we're taught UK English at school. But our languages teaching is so bad it literally taught me nothing, and the English you're the most exposed to on the Internet and through various cultures such as video games, TV series, most of the articles, documentations and so on especially around computer science clearly is the US English which as a result becomes more natural. (Also, I'm not a linguist, but it's funny and I'd be curious to learn about all those little differences on defen(s/c)e, colo(u)r, favo(u)r and so on appeared! (and which variants are actually more common among all the countries using English as a native language or at least an official one, although History bears a lot of importance of the matter).

> As for the formulas, I've used LaTeX to format them which can be done in GitHub by surrounding text in $$ (like t h i s ). If you're not familiar with LaTeX, it's a formatting language specifically used for formulas and is also common in academic papers - for the wiki, it just makes formulas much easier to read.

I know about LaTeX, I had to write an internship report with it, and my CV/resume are produced with it (that's actually not limited to maths at all, and that's what we call a typesetting tool). I never expected it was possible to use it natively into GitHub though. Also, I'm more and more looking at Typst instead of LaTeX. The ecosystem is obviously a little bit less mature because it's a lot younger, but the general tooling and syntax are so much better and more comfortable to use…
(Although if you're author of research papers I heard some journals still asks for some precise formats, and might for example favor `.docs` over PDF produced by this unfortunately…

> As for the questions you mentioned in #1029, only Tristin would probably be able to answer those with any definitive reply since he gets to decide the game direction. One benefit of having it that way is that encourages more variable gameplay - I.e. Do you use lower-level trade routes in order to improve the guild, or do you focus on just maximising mercantile/agility level as quickly as possible?

Well, I've not yet moved to the dedicated F-droid repo (I want to see if the commited apks are addressed first particularly) so I lag behind a bit for versions and only got 1.12.1 12H ago, this morning I was offered… the three little courses from beginning to apprentice while being level 83. That's clearly frustrating.

Another way to see this is that maybe I'll stop caring about these quests and avance them more upon prestiging (make up word, dunno if that works very well!) the ability…

PR #1343, opened 2026-08-11
Improve skills with owned quantity

xRahul:
## Before submitting

## Summary

Add "Owned: N" count to all skill-sheet rows (Mining, Fishing, Woodcutting, Firemaking, Runecrafting, Prayer) and explain the gold dot on skill icons in the Skills tab legend.

N/A

## Changelog

- Add "Owned: N" quantity display to Mining, Fishing, Woodcutting, Firemaking, Runecrafting, and Prayer skill sheet rows
- Explain gold dot icon indicator in Skills tab legend
- Update localization strings across 14 supported locales

## Anything else?

N/A

### DIFF
--- a/app/src/main/kotlin/com/fantasyidler/ui/screen/skills/GatheringSheets.kt
+++ b/app/src/main/kotlin/com/fantasyidler/ui/screen/skills/GatheringSheets.kt
@@ -125,6 +125,7 @@ internal fun MiningSheet(
     petBoostPct: Int = 0,
     xpBonusMult: Float = 1f,
     activeQuests: Map<String, List<QuestIndicator>> = emptyMap(),
+    inventory: Map<String, Int> = emptyMap(),
     onSelect: (String) -> Unit,
 ) {
     val context = LocalContext.current
@@ -170,6 +171,7 @@ internal fun MiningSheet(
                         hasActiveSession = hasActiveSession,
                         isQueueFull      = isQueueFull,
                         questIndicators  = activeQuests["${Skills.MINING}:$key"] ?: emptyList(),
+                        ownedQty         = inventory[key] ?: 0,
                         onClick          = { selectedKey = key },
                     )
                 }
@@ -202,6 +204,7 @@ internal fun WoodcuttingSheet(
     petBoostPct: Int = 0,
     xpBonusMult: Float = 1f,
     activeQuests: Map<String, List<QuestIndicator>> = emptyMap(),
+    inventory: Map<String, Int> = emptyMap(),
     onSelect: (String) -> Unit,
 ) {
     val context = LocalContext.current
@@ -241,6 +244,7 @@ internal fun WoodcuttingSheet(
                         hasActiveSession = hasActiveSession,
                         isQueueFull      = isQueueFull,
                         questIndicators  = activeQuests["${Skills.WOODCUTTING}:${tree.logName}"] ?: emptyList(),
+                        ownedQty         = inventory[tree.logName] ?: 0,
                         onClick          = { selectedKey = key },
                     )
                 }
@@ -273,6 +277,7 @@ internal fun FishingSheet(
     petBoostPct: Int = 0,
     xpBonusMult: Float = 1f,
     activeQuests: Map<String, List<QuestIndicator>> = emptyMap(),
+    inventory: Map<String, Int> = emptyMap(),
     onSelect: (String) -> Unit,
 ) {
     val context = LocalContext.current
@@ -312,6 +317,7 @@ internal fun FishingSheet(
                         hasActiveSession = hasActiveSession,
                         isQueueFull      = isQueueFull,
                         questIndicators  = activeQuests["${Skills.FISHING}:$key"] ?: emptyList(),
+                        ownedQty         = inventory[key] ?: 0,
                         onClick          = { selectedKey = key },
                     )
                 }
@@ -366,8 +372,10 @@ internal fun ActivityRow(
     hasActiveSession: Boolean,
     isQueueFull: Boolean,
     questIndicators: List<QuestIndicator> = emptyList(),
+    ownedQty: Int? = null,
     onClick: () -> Unit,
 ) {
+    val dim = MaterialTheme.colorScheme.onSurface.copy(alpha = 0.38f)
     Row(
         modifier = Modifier
             .fillMaxWidth()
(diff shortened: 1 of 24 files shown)


app/src/main/kotlin/com/fantasyidler/ui/screen/skills/PrayerSheet.kt
```diff
@@ -216,11 +217,19 @@ internal fun PrayerSheet(
...
+                                    style = MaterialTheme.typography.bodySmall,
+                                    color = MaterialTheme.colorScheme.onSurfaceVariant,
+                                )
+                                Spacer(Modifier.width(8.dp))
+                                Text(
+                                    text  = stringResource(R.string.crafting_owned, qty),
+                                    style = MaterialTheme.typography.labelSmall,
+                                    color = if (qty > 0) MaterialTheme.colorScheme.primary else dim,
+                                )
+                            }
```
CiaranGruber: I'm a little curious about this section - how come you chose to keep the "owned" quantity in a row for the prayer sheet but not all the other sheets? If it's not intentional, maybe having it the same as the rest of the sheets could be good?

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/GatheringSheets.kt
```diff
@@ -366,8 +372,10 @@ internal fun ActivityRow(
     hasActiveSession: Boolean,
     isQueueFull: Boolean,
     questIndicators: List<QuestIndicator> = emptyList(),
+    ownedQty: Int? = null,
```
CiaranGruber, line 375: To me it seems like this can just be an `Int` rather than `Int?` since all the calling methods cast it to an int? Small change but may as well fix if that's accurate

app/src/main/res/values-zh-rCN/strings.xml
```diff
@@ -163,4 +163,6 @@
     <string name="settings_backup_success">备份已保存.</string>
     <string name="settings_backup_failed">备份失败.</string>
     <string name="settings_backup_folder_set">备份目录已设置.</string>
+    <string name="skills_xp_each">%1$d XP each</string>
+    <string name="quest_legend_gold_dot">Gold dot on skill icon: has an open daily, weekly, or guild daily quest.</string>
```
CiaranGruber: This is true for the other languages languages as well - to my understanding, these probably shouldn't be added if they are still in English as this obscures them as "Needs translation" on the [translation website](https://translating.tristinbaker.xyz) website - i.e. it means people have to find it in-game and then change it). Android will automatically switch to the default locale when excluding strings so this shouldn't cause any issues

app/src/main/res/values/strings.xml
```diff
@@ -658,6 +657,7 @@
     <string name="quest_legend_guild_daily">Guild daily</string>
     <string name="quest_legend_guild">Guild quest</string>
     <string name="quest_legend_quest">Quest</string>
+    <string name="quest_legend_gold_dot">Gold dot on skill icon: has an open daily, weekly, or guild daily quest.</string>
```
CiaranGruber: Might be good to try and make this more concise so this isn't as long. Maybe just have it say "Timed quest is available"? You could also have it show the full circle without the icon like this if you're worried about it being unclear (though maybe with a larger gold dot):

Image

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/GatheringSheets.kt
```diff
@@ -388,6 +396,13 @@ internal fun ActivityRow(
                 style = MaterialTheme.typography.bodySmall,
                 color = MaterialTheme.colorScheme.onSurfaceVariant,
             )
+            if (ownedQty != null) {
```
CiaranGruber, line 399: Can remove this check if ownedQty is OK removing the null check

CiaranGruber (changes requested):
Overall, looks pretty good and I think the added consistency is good. There are a few changes/considerations I think would be good to implement prior to it getting pulled in which I've commented on to highlight but otherwise thanks for taking the time to contribute

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/GatheringSheets.kt
xRahul: Two of the five ActivityRow callers don't pass ownedQty — AgilitySheet.kt:157 and ThievingSheet.kt:156 use the null default + guard to hide the owned line (no owned-inventory concept there). 
Int = 0 would print 'Owned: 0' on every agility course and thieving NPC. Keeping it nullable.

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/PrayerSheet.kt
xRahul: Not intentional — leftover from splitting the old combined 'XP each • Owned N' string. Moved owned qty onto its own line to match ActivityRow/Firemaking/Runecrafting.

app/src/main/res/values-zh-rCN/strings.xml
xRahul: removed quest_legend_gold_dot from all 15 non-English locales, skills_xp_each from cs/ga/in/zh-rCN where it landed untranslated (kept the 11 translated copies). Default-locale fallback covers the rest.

app/src/main/res/values/strings.xml
xRahul: Shortened to 'Timed quest is available.'

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/GatheringSheets.kt
xRahul: Agility and Thieving don't pass ownedQty — AgilitySheet.kt:157 and ThievingSheet.kt:156 rely on the Int? = null default and this guard to skip the owned line (no owned-inventory concept on those rows). Int = 0 would print "Owned: 0" on every course/NPC. Keeping the null check.

xRahul:
uodated the code

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/GatheringSheets.kt
CiaranGruber: Ah yeah, I see, when I tried it I automatically set a default value which hid the issue but makes sense

CiaranGruber (approved):
Looks good coding-wise and thanks for taking the time to contribute. Will leave it up to @tristinbaker to pull.

Also just to provide a visual of the changes:

image
image

xRahul:
fixed conflicts

PR #1388, opened 2026-08-15
[DO NOT MERGE] Feature/prevent sessions when queue full

Michaelomh:
## Before submitting

## Summary

- [x] Fix in skill section
- [x] Fix in expeditions
- [x] Fix in Dungeons
- [x] Fix infinite Tower
- [x] Fix in Bosses

**Demo**

https://github.com/user-attachments/assets/1b9f2c45-9e75-4225-98c3-3ec4fe70eab8

**Screenshots**
image

## Changelog

- Prevent sessions from being added when queue is full.


xRahul:
The Queue is full is removing information from the screens. We should be able to still see the info on those screens. I feel current popup of Queue is full is good enough. It helps plan things for the future. if we don't see that info, we can't plan until we remove something from the queue

Michaelomh:
@xRahul hey what information is it removing from the screen?

xRahul:
Oh, I just rechecked and seems those places didn't have the xN info already. Should be gtg. I thought we are replacing the info at the right side with these. seems only replacing the add to queue button.

PR #1392, opened 2026-08-15
feat: prevents combat sessions to be created when queue is full

Michaelomh:
## Before submitting

## Summary

- [x] Fix in Dungeons
- [x] Fix infinite Tower
- [x] Fix in Bosses
- [x] Fix in Slayer

**Demo**

https://github.com/user-attachments/assets/d8dcd115-1569-42aa-bd6b-8e402526be38

## Changelog

- Prevent combat sessions from being added when queue is full.

## Anything else?

- There is a bug with Sheets and the AppBanner (something to do with the Z index), but it's something to fix another time.

### DIFF
--- a/app/src/main/kotlin/com/fantasyidler/ui/screen/BossInfoSheet.kt
+++ b/app/src/main/kotlin/com/fantasyidler/ui/screen/BossInfoSheet.kt
@@ -1,6 +1,8 @@
 package com.fantasyidler.ui.screen
 
 import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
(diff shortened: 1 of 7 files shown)


nagi-desuuu:
I guess it supersedes https://github.com/tristinbaker/IdleFantasy/pull/1388 since that pr got closed?

Michaelomh:
Yeah, I decided to close that PR since I realized that gathering/crafting skills was more complicated then I originally thought. Would be raising the PR to prevent non-combat skills from being created when the queue is full. 😃

PR #1394, opened 2026-08-15
Feature/prevent non combat sessions when queue full

Michaelomh:
## Before submitting

## Summary

- [x] Fix in gathering skills section
- [x]  Fix in crafting skills section
- [x]  Fix in prayer skills section
- [x] Fix in expeditions section

**Demo**

https://github.com/user-attachments/assets/f63f79be-c63c-4f28-938b-8ef16b776d92

## Changelog

- Prevent non-combat sessions from being added when queue is full.


app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
```diff
@@ -282,6 +283,7 @@ internal fun CraftSkillSheet(
                         recipe     = recipe,
                         craftState = craftState,
                         context    = context,
+                        isQueueFull=isQueueFull,
```
CiaranGruber: Very small stylistic thing (probably doesn't need fixing unless fixing other issues) though should probably ideally have some spacing between the isQueueFull and the = sign. Not really a problem but good to mention for any future things

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
```diff
@@ -296,6 +298,7 @@ private fun CraftRecipeRow(
     recipe: CraftableRecipe,
     craftState: CraftingUiState,
     context: android.content.Context,
+    isQueueFull: Boolean,
```
CiaranGruber: You should put this as part of the CraftingUiState. The advantage of the CraftingUiState is that if anything updates the CraftingUiState (such as the queue updating) then the view will also update with the correct state - I think this might still happen with isQueueFull being outside of the state but good to keep it all in one place.

You'd probably need to add the appropriate isQueueFull variable [here](https://github.com/Michaelomh/IdleFantasy/blob/375981f37346cfa2d0dbb4188721c72909ac377c/app/src/main/kotlin/com/fantasyidler/ui/viewmodel/CraftingViewModel.kt#L104-124) and [here](https://github.com/Michaelomh/IdleFantasy/blob/375981f37346cfa2d0dbb4188721c72909ac377c/app/src/main/kotlin/com/fantasyidler/ui/viewmodel/CraftingViewModel.kt#L193-L206).

Also, just as a useful tidbit, the thing that actually lets the state know when to update (and subsequent Views) is [this section](https://github.com/Michaelomh/IdleFantasy/blob/375981f37346cfa2d0dbb4188721c72909ac377c/app/src/main/kotlin/com/fantasyidler/ui/viewmodel/CraftingViewModel.kt#L172-L175) of the code. Essentially, if any of those StateFlows change, it causes the function to recalculate the state.

Later Edit: I actually saw you did this correctly for the ExpeditionsViewModel so this would be done the same way for the CraftingUiState

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
```diff
@@ -582,11 +592,30 @@ private fun CraftQuantityContent(
...
+            Button(
+                onClick = { onCraft(qty) },
+                modifier = Modifier.fillMaxWidth(),
+                enabled = !isQueueFull,
+            ) {
+                Text(
+                    if (hasActiveSession) stringResource(R.string.skills_add_to_queue) else stringResource(
+                        R.string.btn_craft
+                    )
+                )
```
CiaranGruber: Small thing though probably good to keep the formatting as it was originally just since it makes it a little easier to review

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
```diff
@@ -582,11 +592,30 @@ private fun CraftQuantityContent(
             }
         }
         Spacer(Modifier.height(20.dp))
-        Button(
-            onClick  = { onCraft(qty) },
-            modifier = Modifier.fillMaxWidth(),
-        ) {
-            Text(if (hasActiveSession) stringResource(R.string.skills_add_to_queue) else stringResource(R.string.btn_craft))
+        val queueFullMessage = stringResource(R.string.snackbar_queue_full)
```
CiaranGruber: It would usually be better to have this stringResource be calculated in the same line (L615) for consistency and also keeps everything together which makes it easier to debug in the future. This is with the exception of if the same string was to be used in many places though I haven't seen that in the codebase yet

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
```diff
@@ -582,11 +592,30 @@ private fun CraftQuantityContent(
...
+            if (isQueueFull) {
+                Box(
+                    modifier = Modifier
+                        .matchParentSize()
+                        .clickable(
+                            interactionSource = remember { MutableInteractionSource() },
+                            indication = null,
+                            onClick = { AppBannerCenter.enqueue(queueFullMessage) },
+                        ),
+                )
```
CiaranGruber: I like this UI design - it means people will get less confused when they try to click the button. Perhaps a simpler solution however is just to change the text on the button to "Queue is full" while it's disabled? This is just opinion however so feel free to keep what you think is better and Tristin can keep what he prefers

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
```diff
@@ -239,72 +251,89 @@ internal fun FiremakingSheet(
                     .verticalScroll(detailScrollState)
                     .imePadding(),
             ) {
-                TextButton(onClick = { selectedKey = null }, modifier = Modifier.padding(start = 4.dp)) {
+                TextButton(
+                    onClick = { selectedKey = null },
+                    modifier = Modifier.padding(start = 4.dp)
+                ) {
```
CiaranGruber: Like previous, would be better to keep the code styles as-is to make it easier for review

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
```diff
@@ -239,72 +251,89 @@ internal fun FiremakingSheet(
...
                     Text(stringResource(R.string.btn_back_arrow))
                 }
                 Text(
-                    text     = GameStrings.itemName(context, key),
-                    style    = MaterialTheme.typography.titleMedium,
+                    text = GameStrings.itemName(context, key),
+                    style = MaterialTheme.typography.titleMedium,
```
CiaranGruber: From my time working with the codebase it seems to be standard to align the equals symbols so probably better to keep as it was

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
```diff
@@ -239,72 +251,89 @@ internal fun FiremakingSheet(
...
                 ) { Text(stringResource(R.string.firemaking_burn)) }
+                if (isQueueFull) {
+                    Box(
+                        modifier = Modifier
+                            .matchParentSize()
+                            .clickable(
+                                interactionSource = remember { MutableInteractionSource() },
+                                indication        = null,
+                                onClick           = { AppBannerCenter.enqueue(queueFullMessage) },
+                            ),
```
CiaranGruber: Similar comment to earlier about potential options - just marking so it isn't missed

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
```diff
@@ -239,72 +251,89 @@ internal fun FiremakingSheet(
...
+            val logPerMs = perLogMs[key]?.takeIf { it > 0 } ?: (sessionDurationMs / 60)
+            if (logPerMs > 0) {
+                Text(
+                    text     = "~${(qty.toLong() * logPerMs).formatDurationMs()}",
+                    style    = MaterialTheme.typography.bodySmall,
+                    color    = MaterialTheme.colorScheme.onSurfaceVariant,
+                    modifier = Modifier.padding(start = 16.dp, end = 16.dp, bottom = 8.dp),
+                )
+            }
+            val queueFullMessage = stringResource(R.string.snackbar_queue_full)
```
CiaranGruber: Like previous, would be good to just have it with where it's used to keep it together

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
```diff
@@ -239,72 +251,89 @@ internal fun FiremakingSheet(
...
-                    text     = stringResource(R.string.firemaking_ashes_owned, ashOwned),
-                    style    = MaterialTheme.typography.labelSmall,
-                    color    = if (ashOwned > 0) MaterialTheme.colorScheme.primary else dim,
+                    text = stringResource(R.string.firemaking_ashes_owned, ashOwned),
+                    style = MaterialTheme.typography.labelSmall,
+                    color = if (ashOwned > 0) MaterialTheme.colorScheme.primary else dim,
                     modifier = Modifier.padding(horizontal = 16.dp, vertical = 4.dp),
                 )
                 Spacer(Modifier.height(12.dp))
+            }
```
CiaranGruber: This looks like it was inadvertently added which caused all the changes below (which is mostly just changes to padding but this makes it hard to review) and also means the code below no longer part of the Row (which it probably should be unless you have a reason for removing).

Should probably delete this and then readd the } at the bottom of the Firemaking sheet somewhere [here](https://github.com/Michaelomh/IdleFantasy/blob/375981f37346cfa2d0dbb4188721c72909ac377c/app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt#L338)

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/GatheringSheets.kt
```diff
@@ -450,10 +465,17 @@ internal fun ActivityDetailDialog(
             }
         },
         confirmButton = {
+            val queueFullMessage = stringResource(R.string.snackbar_queue_full)
```
CiaranGruber, line 469: Based off previous comments, should probably be with the line which uses it (L472)

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/PrayerSheet.kt
```diff
@@ -227,14 +228,26 @@ internal fun PrayerSheet(
                                     color = MaterialTheme.colorScheme.onSurfaceVariant,
                                 )
                             }
-                            Text(
-                                text  = stringResource(R.string.crafting_owned, qty),
-                                style = MaterialTheme.typography.labelSmall,
-                                color = if (qty > 0) MaterialTheme.colorScheme.primary else dim,
-                            )
+                            Column(horizontalAlignment = Alignment.End) {
+                                Text(text = " ", style = MaterialTheme.typography.bodyLarge)
```
CiaranGruber: Not sure if this was intentional? To me it feels like the text is better vertically centred?

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/PrayerSheet.kt
```diff
@@ -227,14 +228,26 @@ internal fun PrayerSheet(
...
+                                        color = dim,
+                                    )
+                                }
+                            }
                         }
-                        HorizontalDivider(modifier = Modifier.padding(horizontal = 16.dp))
+
+
                     }
+                    HorizontalDivider(modifier = Modifier.padding(horizontal = 16.dp))
```
CiaranGruber: This should be moved one line/block up else it is effectively missing from the bones

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/PrayerSheet.kt
```diff
@@ -329,21 +342,36 @@ internal fun PrayerSheet(
                 )
             }
 
-            Button(
-                onClick  = { onStart(selectedKey!!, qty) },
-                enabled  = !isStarting && qty > 0 && maxQty > 0,
+            val queueFullMessage = stringResource(R.string.snackbar_queue_full)
```
CiaranGruber: Should be with the place it's used to keep things together

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/PrayerSheet.kt
```diff
@@ -329,21 +342,36 @@ internal fun PrayerSheet(
...
+            Box(
                 modifier = Modifier
                     .fillMaxWidth()
                     .padding(horizontal = 16.dp),
```
CiaranGruber, line 346: Nice fix putting this as part of the box. If you want, it actually would probably be a good fix if this whole region instead used a [column like in the CraftSkillSheet](https://github.com/Michaelomh/IdleFantasy/blob/375981f37346cfa2d0dbb4188721c72909ac377c/app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt#L459-L466), with the appropriate horizontal/vertical padding and proper 12.dp spacers rather than each element defining its own padding but that's up to you if you want to fix that - feel free to leave as is

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/RunecraftingSheet.kt
```diff
@@ -154,93 +156,111 @@ internal fun RunecraftingSheet(
...
+                                    color = dim,
+                                )
+                            }
                         }
-                        HorizontalDivider(modifier = Modifier.padding(horizontal = 16.dp))
                     }
+                    HorizontalDivider(modifier = Modifier.padding(horizontal = 16.dp))
                 }
             }
+            }
```
CiaranGruber: It'd be good if you can fix the padding in these lines to match what it was previously as this makes it easier to review as GitHub won't mark it as changes needing review provided you haven't added anything in that actually does need reviewing (though they appear to be the same from comparing)

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/RunecraftingSheet.kt
```diff
@@ -384,15 +404,31 @@ internal fun RunecraftingSheet(
                 }
             }
 
-            Button(
-                onClick  = { onStart(selectedKey!!, qty, selectedAshKey) },
-                enabled  = !isStarting && qty > 0 && maxQty > 0,
+            val queueFullMessage = stringResource(R.string.snackbar_queue_full)
```
CiaranGruber: Just marking like previous things - good to keep together with where it's used

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/RunecraftingSheet.kt
```diff
@@ -384,15 +404,31 @@ internal fun RunecraftingSheet(
...
+            Box(
                 modifier = Modifier
                     .fillMaxWidth()
                     .padding(horizontal = 16.dp),
```
CiaranGruber, line 344: Like mentioned in the PrayerSheet, could be nice if you added a Column so this padding didn't need to be included in every element though up to you if you would like to do this

app/src/main/kotlin/com/fantasyidler/ui/screen/ExpeditionsScreen.kt
```diff
@@ -134,11 +138,25 @@ private fun SkillingDungeonCard(
                     )
                 }
                 Spacer(modifier = Modifier.width(8.dp))
-                Button(
-                    onClick = onExplore,
-                    enabled = item.isAccessible,
-                ) {
-                    Text(stringResource(R.string.expedition_explore_button))
+                val queueFullMessage = stringResource(R.string.snackbar_queue_full)
```
CiaranGruber, line 155: Same as previous comments about keeping it with where it's used

app/src/main/kotlin/com/fantasyidler/ui/screen/MercantileScreen.kt
```diff
@@ -210,6 +212,7 @@ private fun TradeRouteRow(
     val costStr   = route.coinCost.toLong().formatCoins()
     val minReturn = (route.coinRanges.values.minOf { it.min } * 60L * coinReturnMult.toDouble()).toLong()
     val maxReturn = (route.coinRanges.values.maxOf { it.max } * 60L * coinReturnMult.toDouble()).toLong()
+    val queueFullMessage = stringResource(R.string.snackbar_queue_full)
```
CiaranGruber, line 215: Marking as per previous comments

CiaranGruber (changes requested):
Overall nice job, I like what you've done. There are a lot of changes to make however they're all mainly just minor fixes that I noticed as I was reviewing. Btw, I'm just doing a review so if you intentionally did something that you think I missed (I can make mistakes as well) or you prefer the approach you took, feel free to mention - you don't need to apply everything I suggest though it would be good to know if something was intentional.

Thanks for taking the time to contribute and once you've applied/discussed the fixes, then I think it should be good to go from a code quality standpoint (will leave for Tristin to pull after)

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
Michaelomh: updated

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
Michaelomh: Nice, thanks for the tip! Honestly, still learning kotlin and was just utilizing what I understand from other languages to come out with something (hopefully) semi decent. 🙏 I've completely forgotten about the ViewModel here. 

Updated.

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
Michaelomh: wow, i'm surprised that this project don't have a prettier. Something that I can look into in the future. Updated.

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
Michaelomh: Correct me if i'm wrong, but Line 615 is inside an onclick, which is not a composable context and thus, stringResource cannot be called there and needs to be hoisted to a composable context first. 

either way, the need for queueFullMessage is no longer needed due to this comment - https://github.com/tristinbaker/IdleFantasy/pull/1394/changes#r3789647236. I don't want to have an appBanner to appear while the button is disabled. If the button already says "Queue is full". It's just reductive.

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
Michaelomh: I've updated the text and remove the need for an appBanner. Doesn't make sense to me to show the appBanner with the exact same text.

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
Michaelomh: updated

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
Michaelomh: we really need a prettier 😄

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
Michaelomh: updated

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
Michaelomh: updated. These places needs refactoring 🙈

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
Michaelomh: Thanks for the catch. Updated and verified.

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/PrayerSheet.kt
Michaelomh: Fixed this. I think it used to be right, but the merge didn't go as expected. Thanks for catching this!

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/PrayerSheet.kt
Michaelomh: Fixed.

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/PrayerSheet.kt
Michaelomh: Not required anymore since we are not showing appBanner when queue is full.

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/PrayerSheet.kt
Michaelomh: Not required anymore since we are not showing appBanner when queue is full.

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/RunecraftingSheet.kt
Michaelomh: More reasons for a prettier! Fixed.

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/RunecraftingSheet.kt
Michaelomh: Not required anymore since we are not showing appBanner when queue is full.

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/RunecraftingSheet.kt
Michaelomh: See the other comments about `@Composable invocations can only happen from the context of a @Composable function`

Image

app/src/main/kotlin/com/fantasyidler/ui/screen/ExpeditionsScreen.kt
Michaelomh: See the other comments about `@Composable invocations can only happen from the context of a @Composable function`

Image

app/src/main/kotlin/com/fantasyidler/ui/screen/MercantileScreen.kt
Michaelomh: Hmmm, I don't think is the case. Since the onClick is not in a composable context and thus, stringResource cannot be called there and needs to be hoisted to a composable context first.

see: 
Image

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/GatheringSheets.kt
Michaelomh: See the other comments about `@Composable invocations can only happen from the context of a @Composable function`

Image

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
CiaranGruber: Hey no problem, I completely get it and that's one of the advantages of the review is that there might be something I might be aware of that you didn't know just from having worked with it for a bit

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
CiaranGruber: Yeah, I'm not too sure the process so I've mainly just kept to how Tristin does it but a prettier setup could be good - you could maybe message Tristin on the #development chat about setting up some standards?

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
CiaranGruber: Ah no, you're absolutely right, that's my bad. I was mainly looking at it from the perspective of being in the view but you are right in that the onClick takes it out off the Composable context.

For future reference btw (although it not be relevant anymore as you mentioned you're sticking to not having the banner), if it is outside of the Composable context, there's another pattern to retrieve strings where you do the following:

```kotlin

val context = LocalContext.current
context.getString(R.string.snackbar_queue_full)
```

This would probably be the best approach in situations like these or if you need to define it in the ViewModel for example

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
```diff
@@ -208,11 +208,22 @@ internal fun FiremakingSheet(
...
+                            Column(horizontalAlignment = Alignment.End) {
+                                Text(text = " ", style = MaterialTheme.typography.bodyLarge)
+                                Text(
+                                    text = "$logsOwned ${stringResource(R.string.firemaking_logs_in_inventory)}",
+                                    style = MaterialTheme.typography.labelSmall,
+                                    color = MaterialTheme.colorScheme.onSurfaceVariant.copy(alpha = rowAlpha),
+                                )
+                                if (isQueueFull) {
+                                    Text(
+                                        text = context.getString(R.string.snackbar_queue_full),
```
CiaranGruber: This is the good method for getting strings outside of Composable context that I mentioned as a response in the previous review. Would recommend using stringResource where possible (like here as the text= in Text is considered Composable context) but that's nothing major

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/GatheringSheets.kt
CiaranGruber: Yep, that was a good catch. Can see [here](https://github.com/tristinbaker/IdleFantasy/pull/1394#discussion_r3791643935) for previous comments

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/RunecraftingSheet.kt
```diff
@@ -209,7 +211,7 @@ internal fun RunecraftingSheet(
                                 .clickable { selectedKey = key }
                                 .padding(horizontal = 16.dp, vertical = 12.dp),
                             horizontalArrangement = Arrangement.SpaceBetween,
-                            verticalAlignment = Alignment.CenterVertically,
+                            verticalAlignment = Alignment.Top,
```
CiaranGruber, line 154: Just curious why this was changed? That's all good though since it didn't seem to affect anything major

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/RunecraftingSheet.kt
```diff
@@ -231,11 +233,21 @@ internal fun RunecraftingSheet(
...
+                            Column(horizontalAlignment = Alignment.End) {
+                                Text(text = " ", style = MaterialTheme.typography.bodyLarge)
+                                Text(
+                                    text  = stringResource(R.string.btn_select),
+                                    style = MaterialTheme.typography.labelMedium,
+                                    color = MaterialTheme.colorScheme.primary,
+                                )
+                                if (isQueueFull) {
+                                    Text(
+                                        text  = context.getString(R.string.snackbar_queue_full),
```
CiaranGruber: Another place where it would be good to put as a string resource but all good

app/src/main/kotlin/com/fantasyidler/ui/screen/ExpeditionsScreen.kt
CiaranGruber: Yep, that was a good catch. Can see [here](https://github.com/tristinbaker/IdleFantasy/pull/1394#discussion_r3791643935) for previous comments

app/src/main/kotlin/com/fantasyidler/ui/screen/MercantileScreen.kt
CiaranGruber: Yep, that was a good catch. Can see [here](https://github.com/tristinbaker/IdleFantasy/pull/1394#discussion_r3791643935) for previous comments

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/RunecraftingSheet.kt
CiaranGruber: I was actually more referring to the padding rather than the queueFullMessage in this case though as mentioned, it is fine to leave as is - just more us to refactor at some point in the future. The comment I was referencing was [here](https://github.com/tristinbaker/IdleFantasy/pull/1394#discussion_r3789719674).

app/src/main/kotlin/com/fantasyidler/ui/screen/ExpeditionsScreen.kt
```diff
@@ -134,11 +138,25 @@ private fun SkillingDungeonCard(
...
+                        Text(stringResource(R.string.expedition_explore_button))
+                    }
+                    if (item.isAccessible && isQueueFull) {
+                        Box(
+                            modifier = Modifier
+                                .matchParentSize()
+                                .clickable(
+                                    interactionSource = remember { MutableInteractionSource() },
+                                    indication        = null,
+                                    onClick           = { AppBannerCenter.enqueue(queueFullMessage) },
```
CiaranGruber, line 170: If you wanted, could switch this to the recommended context.getString(R.string.queueFullMessage).

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/GatheringSheets.kt
```diff
@@ -450,10 +465,17 @@ internal fun ActivityDetailDialog(
...
             Button(
-                onClick = { onConfirm(); onDismiss() },
+                onClick = {
+                    if (isQueueFull) {
+                        AppBannerCenter.enqueue(queueFullMessage)
```
CiaranGruber, line 473: Just marking another place where context.getString(R.string.snackbar_queue_full) could be placed if you wanted to fix

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/GatheringSheets.kt
```diff
@@ -379,10 +380,20 @@ internal fun ActivityRow(
...
+    val queueFullMessage = stringResource(R.string.snackbar_queue_full)
     Row(
         modifier = Modifier
             .fillMaxWidth()
-            .clickable(enabled = !isStarting, onClick = onClick)
+            .clickable(
+                enabled = !isStarting,
+                onClick = {
+                    if (isQueueFull) {
+                        AppBannerCenter.enqueue(queueFullMessage)
```
CiaranGruber, line 392: Just marking another place where context.getString(R.string.snackbar_queue_full) could be placed if you wanted to fix

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
```diff
@@ -557,7 +562,10 @@ private fun CraftQuantityContent(
                 val selectedAsh = state.herbloreAshKey
                 (listOf(null) + availableAshes).forEach { ashKey ->
                     Row(
-                        modifier          = Modifier.fillMaxWidth().clickable { onSetAsh(ashKey) }.padding(vertical = 4.dp),
+                        modifier          = Modifier
+                            .fillMaxWidth()
+                            .clickable { onSetAsh(ashKey) }
+                            .padding(vertical = 4.dp),
```
CiaranGruber, line 568: Good to keep original style in general but very minor and as you mentioned - a prettier setup would probably solve that issue (same for code further above)

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
```diff
@@ -416,6 +413,13 @@ private fun CraftRecipeRow(
                     color = dim,
                 )
             }
+            if (craftState.isQueueFull) {
+                Text(
+                    text  = context.getString(R.string.snackbar_queue_full),
```
CiaranGruber: Just marking a place where you could switch to stringResource() if you wanted as text = is still composable context (unlike onClick which executes in a Unit (i.e. lambda))

CiaranGruber (approved):
Nice job and I think it should be all good to go. Mentioned it before but thanks for contributing and hopefully you feel the review was helpful for you as well to get familiar with Kotlin.

There were a few minor points I've raised however these are in no way blocking issues so if you would like to fix, that'd be good (keeps the codebase that little bit tidier/consistent) but otherwise it's fine - code gets refactored in the future anyway

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
Michaelomh: ah thanks, updated

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
Michaelomh: updated

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/CraftSkillSheet.kt
Michaelomh: updated

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/RunecraftingSheet.kt
Michaelomh: updated

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/RunecraftingSheet.kt
Michaelomh: It's because the new "Queue is full" Text was added to that Column. Top alignment stops the row re-centering oddly with the extra line. I thought this would work similarly to CSS Flexbox/Grid (it might be, but I don't have time to look into this further), but I guess it's slightly different. but it's the same result.

app/src/main/kotlin/com/fantasyidler/ui/screen/skills/FiremakingSheet.kt
Michaelomh: fixed

Michaelomh:
Hey @CiaranGruber @tristinbaker, I've also added a discussion on using prettiers/formatter here - https://github.com/tristinbaker/IdleFantasy/discussions/1405 and  https://github.com/tristinbaker/IdleFantasy/discussions/1406

puggan:
Related issues/discussions:
#1359 & #1251

PR #1440, opened 2026-08-17
WIKI: updating the contribution guidelines and information

wallsocket-boop:
## Before submitting

## Summary

I am in the process of updating the contribution guidelines on the wiki, to add more information, and to make some things clearer.

I would appreciate any feedback on it so far, and in to the future as to what to exclude / include.

Currently planned:
- Add more information on adding support for translation strings.

## Changelog

- updated the wiki building guidelines

### DIFF
--- a/wiki/templates/contributing/getting_started_wiki.md
+++ b/wiki/templates/contributing/getting_started_wiki.md
@@ -67,7 +67,7 @@ To build the wiki, you'll need to set up an appropriate Python environment in th
 ```bash
 pip install -r wiki/requirements.txt
 ```
-Make sure you've installed all packages shown in `wiki/requirements.txt`. You can do this by running the above command from the root of the repository.
+Make sure you've installed all packages shown in `wiki/requirements.txt`. You can do this by running the above command in root folder for the repository.
 
 ### 3. Building the wiki 
 ```bash
--- a/wiki/templates/contributing/getting_started_wiki.md
+++ b/wiki/templates/contributing/getting_started_wiki.md
@@ -48,14 +48,14 @@ Whenever we refer to static vs dynamically-generated pages, this refers not to t
 
 For more information, see {page_types_link}.
 
-## Building/Compiling the Wiki
+## Building from source
 
 To build the wiki, you'll need to set up an appropriate Python environment in the root directory. For information about setting up virtual environments, see [this tutorial](https://www.geeksforgeeks.org/python/create-virtual-environment-using-venv-python/). 
 *Make sure to call your virtual environment `.venv` to have it be gitignored).*
 
 ### 1. Activate the virtual environment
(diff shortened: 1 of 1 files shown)


wiki/templates/contributing/getting_started_wiki.md
```diff
@@ -50,25 +50,44 @@ For more information, see {page_types_link}.
...
+python -m wiki.src validity
+```
+You can use the above command to validate the wiki and perform several tests that can help you pick up errors.
 
-`python -m wiki.src validity` is the next most common command which you can use to validate the wiki and perform several tests that can help you pick up errors
+### 5. Additional commands, command usage & help
+```bash
+python -m wiki.src -h
+```
+You can use the above command to display a list of what else you can do.
```
CiaranGruber: Maybe this could be put this under a heading like `### Usage` with other headings `#### Compiling the wiki` `#### Validating the wiki`, etc (without numbers as you can do any command in order)?

This was made rather quickly so a better restructure of the whole Building/Compiling section could have it in two sections: `## Building from source` and `## Usage` with `## Usage` explaining all the things like validation, write-html & additional commands

CiaranGruber (commented):
Nice, I like what you've done, I think it makes the documentation much clearer. If anything, I think you could have done a bit more potentially since splitting the sections into `building from source` and `usage` is standard for this sort of documentation and could be worthwhile to do as an improvement.

I think this should be good to pull in as is but if you wanted to make the aforementioned changes then you could do so

wiki/templates/contributing/getting_started_wiki.md
```diff
@@ -69,26 +69,29 @@ pip install -r wiki/requirements.txt
 ```
 Make sure you've installed all packages shown in `wiki/requirements.txt`. You can do this by running the above command in root folder for the repository.
 
-### 3. Building the wiki 
+### 3. Usage
```
wallsocket-boop, line 72: I've kept this as 3. because it should only happen after 1. and 2., however I've left the sections within it as open because they are more optional in their ordering.

wiki/templates/contributing/getting_started_wiki.md
```diff
@@ -109,7 +112,7 @@ If you want to make a number of new pages or significant changes (especially dyn
 
 Additionally, you might notice a number of #Todo items in the wiki code. These are also great places to look for things that need doing that we maybe just haven't got around to sorting out yet.
 
-## Pull request guides
+### Pull request guides
```
wallsocket-boop, line 115: I've put this under Contributing process, as I feel it is relevant.

wallsocket-boop:
I'm going to reduce the scope this pr to just this, to reduce the burden of scope and allow preliminary alterations to be created sooner than later, then mark it as ready for review.

Unless you think there is anything else that can be easily added into this pr, I think this is ready for review.

Thank you for the feedback.

PR #1447, opened 2026-08-17
Fix profile tab resetting when navigating away and back

Michaelomh:
## Before submitting

## Summary

Navigating away from the Profile screen and back reset the selected tab to the first tab instead of restoring the tab the user was previously on. It doesn't seem to be the expected behaviour, since other tabs are not doing the same. 

**Before PR (ignore the stats section, I'm doesn't belong in this PR)**

https://github.com/user-attachments/assets/da65eabf-9615-4309-b379-dc98c0905bd9

**After PR**

https://github.com/user-attachments/assets/7537e6f2-6a6c-4668-aace-e2256919da18

## Changelog

- Restore Profile screen state (including tab selection) when navigating back to it.

## Anything else?

### DIFF
--- a/app/src/main/kotlin/com/fantasyidler/ui/navigation/AppNavigation.kt
+++ b/app/src/main/kotlin/com/fantasyidler/ui/navigation/AppNavigation.kt
@@ -140,7 +140,7 @@ fun AppNavigation(
                                         saveState = true
                                     }
                                     launchSingleTop = true
-                                    restoreState = !isHome && screen !is Screen.Profile
+                                    restoreState = !isHome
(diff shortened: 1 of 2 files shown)


puggan:
I see why profile -> home -> profile should show the same tab as before,
but if you are on profile and click on profile again, I'd expect the profile view to reset, in your after video it just ignores that click.

If a user click on profile when you already are on the profile, they probably expect something to happen, and resetting the profile view as in the before video seams like an ok action, the after version feels like we just think the user is an idiot.

(just a personal opinion, from a player)

PR #1458, opened 2026-08-18
i18n: complete zh-CN translations for v1.13.10

LawrenZzz:
- Translate 3 new strings (switch character, daily reset time)
- Fix 32 previously untranslated strings
- Remove 68 extra translations not in default locale
- Clean up all untranslated markers

## Before submitting

- [ ] I've tested any changes with the appropriate tests (eg. Unit tests, validation scripts)
- [ ] I've pulled and merged the latest changes from the repository

## Changelog

-

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -26,10 +26,10 @@ android {
 
     signingConfigs {
         create("release") {
-            storeFile     = file("${System.getProperty("user.home")}/.android/defide-release.jks")
-            storePassword = System.getenv("DEFIDE_STORE_PASSWORD") ?: ""
-            keyAlias      = "defide"
-            keyPassword   = System.getenv("DEFIDE_KEY_PASSWORD") ?: ""
+            storeFile     = file("${System.getProperty("user.home")}/.android/idlefantasy-release.jks")
+            storePassword = System.getenv("IDLEFANTASY_STORE_PASSWORD") ?: ""
+            keyAlias      = "idlefantasy"
+            keyPassword   = System.getenv("IDLEFANTASY_KEY_PASSWORD") ?: ""
         }
     }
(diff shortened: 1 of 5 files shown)


app/src/main/res/values-zh-rCN/strings_quests.xml
```diff
@@ -508,8 +508,8 @@
     <string name="daily_verb_farming">收获</string>
 
     <!-- Quest objective format strings -->
-    <string name="quest_obj_gather">%1$s %2$d %3$s</string> <!-- untranslated -->
```
Michaelomh, line 511: Looking at the lint failures, I believe this is the reason why the lint is failing.

app/build.gradle.kts
```diff
@@ -26,10 +26,10 @@ android {
 
     signingConfigs {
         create("release") {
-            storeFile     = file("${System.getProperty("user.home")}/.android/defide-release.jks")
```
Michaelomh, line 29: Why did you update this? Was this some merging issues?

Michaelomh (changes requested):
Hey @LawrenZzz, thanks for contributing to the translations in Mandarin. I've pointed out what is causing the build to fail. Please take a look at them.

app/build.gradle.kts
tristinbaker: Agreed, the changes to this file will break the build process. As is, this PR is not able to be merged.

tristinbaker:
Thanks for completing the zh-CN translations and cleaning out the stale keys! Merged. One note: the signing config rename in app/build.gradle.kts was reverted right after the merge since it pointed at a keystore path and environment variables that don't match the release setup. The translations ship in v1.14.0.

PR #1479, opened 2026-08-19
Make prayer cape consistently boost church blessings

### LINKED ISSUE #1491: Prayer cape doesn't apply bonus to all boosts
### Before submitting

- [x] I am on the latest available version (see \#516)

### App Version

1.13.11

### Android Version

17

### Device

Pixel 9a

### What happened?

The prayer cape says it adds a 10% boost bonus, but it doesn't seem to.

### Steps to reproduce

1. Open the app
2. Have the prayer cape equipped
3. Go to the church screen
4. Buy a blessing xp/coins boost
5. Queue a session yielding xp/coins
6. Collect the session and observe the boost bonus from the cape wasn't applied

### Expected behavior

The blessing boost should be increased by the bonus from the cape (and prestige levels affecting the cape)

Every time

Blessing boost application

### Relevant game state

Prayer cape equipped, blessing boost purchased

### Logs or crash trace

```shell

```

### Anything else?

I browsed the code and saw the bonus from the cape is only applied to the defense boost and not everywhere.
I've made a pull request with changes I think would fix this, making the cape bonus apply consistently to all tree kinds of blessing boosts

lucasmdjl:
## Before submitting

## Summary

This PR makes the prayer cape bonus apply consistently to the church blessing calculations, centralizing it in ChurchRepository instead of relying on callers multiplying the blessing by the cape bonus (which previously was only done in some cases and only for the defense blessing).
I believe this change makes it worthwhile to prestige the prayer skill, as there's little reason to do it otherwise.

## Changelog

- Make prayer cape consistently boost church blessings
- Update xp multiplier format in home and profile screens to be in line with church formatting

## Anything else?

The changes make prayer prestige potentially influence more strongly the xp gained in the bone altar when the prayer cape is equipped and the xp blessing is active. However, I think this should be the intended behaviour, the same way that for other skills the prestige would interact with the church blessing (and thus with the prayer prestige).

Closes #1491

### DIFF
--- a/app/src/main/kotlin/com/fantasyidler/repository/ChurchRepository.kt
+++ b/app/src/main/kotlin/com/fantasyidler/repository/ChurchRepository.kt
@@ -2,6 +2,8 @@ package com.fantasyidler.repository
 
 import com.fantasyidler.data.json.BlessingData
 import com.fantasyidler.data.json.BlessingType
+import com.fantasyidler.data.json.EquipmentData
+import com.fantasyidler.data.model.EquipSlot
 import com.fantasyidler.data.model.PlayerFlags
 import com.fantasyidler.data.model.Skills
 import javax.inject.Inject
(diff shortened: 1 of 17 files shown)


tristinbaker:
Thanks so much for digging into this and for the fix! The upcoming 1.14 branch has diverged quite a bit from main in how cape bonuses are computed, so rather than merging I've implemented the fix directly on that branch: the prayer cape now consistently strengthens all blessing types, and the church screen and stats bar show the boosted values. You're credited in the changelog.

A couple of notes on the parts I didn't carry over:
• The session-collection refactor looks like a genuinely nice cleanup, but I'd rather take it as its own PR after 1.14 ships so it can be reviewed against the new code. Would happily welcome that one!
• The bone altar change shifts current balance, so I've kept the existing behavior there for now.

Closing this since the fix lands in v1.14.0. Thanks again for the contribution!

Issue #1496, opened 2026-08-21
Coin Drake 	Mercantile (1/1000 per craft), should be per session


OlMon:
### Before submitting

- [x] I searched existing issues and this hasn't been reported
- [x] This relates to the latest version of the game
- [ ] _(Optional)_ I would like to fix this myself (Thanks! See [how to get started](https://idlefantasy.tristinbaker.xyz/getting_started_wiki.html))

### App Version

1.13.11

### Page URL

https://idlefantasy.tristinbaker.xyz/Pets.html

### What's incorrect?

I am not sure if this is really a wiki bug, the wiki reflects the current state and the state is wrong, or it is as intended.

Looking at Coin Drake, the drop chance is listed as "per craft", this does not make a lot of sense. It should probably be "per session". The chance is also really low in comparison to other skills. 

As i said I am not sure where the bug is, that is why i did not just fix the json and would like if some one more familiar with the codebase would look into it.

### Expected Text

I would expect:
Coin Drake 	Mercantile (6% per session)

When the code reflects that.

### Which part of the game does this relate to?

Coin Drake Pet drop chance.

ReflexS777:
Its the only pet im still missing, doing lots of mercantile, and it won't drop.
Im max prestige mercantile, so i did alot of it and it still hasn't dropped.

All the other skill pets dropped fairly easy and normally while playing.

OlMon:
I am in the same boat, that is why i checked what is on the wiki....

tristinbaker:
Good instinct, this was a game bug and not just a wiki typo. Trade routes started directly from the Mercantile screen never rolled the Coin Drake at all; only routes added through the queue could award it. That is why it never dropped for you no matter how much Mercantile you did. The same gap also affected crafting pets and the agility pet for directly started sessions.

This will be fixed in v1.14.0, and the drop text now correctly reads a roughly 6% chance per session, the same as the gathering pets. Thanks for the report, and thanks ben-dev-su for the fix!

PR #1501, opened 2026-08-21
fix: 'Coin Drake' pet drops for mercantile trade routes

ben-dev-su:
## Before submitting

## Summary

Fixes Coin Drake drops for Mercantile trade routes started directly.

Directly started routes did not pass the Mercantile pet ID or drop
chance to `MercantileSimulator`. Queued routes already passed these
values correctly.

Might relate to #1496

## Changelog

- Fixed Coin Drake drops for directly started Mercantile trade routes


tristinbaker:
Thanks for tracking this down! You found exactly the right spot: directly started trade routes never rolled the pet at all. Merging this, and the upcoming release also extends the same fix to directly started crafts and agility courses, which had the same gap. Fix ships in v1.14.0.

PR #1628, opened 2026-08-30
Add year to duration strings. 250mo -> 20y 7mo

redjard:
Currently formatDurationMs tops out at months. However if a player has a large XP or prayer timer, or is (hopefully not) starting a very long session, then a time in years is a lot easier to understand than one in months.

Screenshot_20260830-190711_Idle Fantasy

Before:  
Screenshot_20260830-185455_Idle Fantasy

After:  
Screenshot_20260830-185425_Idle Fantasy

I have successfully built the apps and tests. The screenshots are from my phone, where it installed and worked correctly.

## Before submitting

## Changelog

- Very long durations now display with years instead of huge month counts (e.g. 250mo 3w 1d 8h 7m becomes 20y 7mo 1w 5d 8h 7m)

### DIFF
--- a/app/src/main/kotlin/com/fantasyidler/util/Extensions.kt
+++ b/app/src/main/kotlin/com/fantasyidler/util/Extensions.kt
@@ -122,12 +122,14 @@ internal fun Long.formatDurationMs(unitString: (Int, Long) -> String): String {
     var rem = totalSeconds / 60
     if (rem == 0L) return unitString(R.string.duration_seconds, totalSeconds)
     val minutesPerDay = 24L * 60
-    val months = rem / (30 * minutesPerDay); rem %= 30 * minutesPerDay
-    val weeks  = rem / (7 * minutesPerDay);  rem %= 7 * minutesPerDay
-    val days   = rem / minutesPerDay;        rem %= minutesPerDay
+    val years  = rem / (365 * minutesPerDay); rem %= 365 * minutesPerDay  // this does allow 1y 12mo 4 days, which is acceptable
+    val months = rem / (30 * minutesPerDay);  rem %= 30 * minutesPerDay
+    val weeks  = rem / (7 * minutesPerDay);   rem %= 7 * minutesPerDay
+    val days   = rem / minutesPerDay;         rem %= minutesPerDay
     val hours  = rem / 60
     val minutes = rem % 60
     return buildList {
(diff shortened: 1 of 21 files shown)


app/src/main/kotlin/com/fantasyidler/util/Extensions.kt
```diff
@@ -122,12 +122,14 @@ internal fun Long.formatDurationMs(unitString: (Int, Long) -> String): String {
     var rem = totalSeconds / 60
     if (rem == 0L) return unitString(R.string.duration_seconds, totalSeconds)
     val minutesPerDay = 24L * 60
-    val months = rem / (30 * minutesPerDay); rem %= 30 * minutesPerDay
-    val weeks  = rem / (7 * minutesPerDay);  rem %= 7 * minutesPerDay
-    val days   = rem / minutesPerDay;        rem %= minutesPerDay
+    val years  = rem / (365 * minutesPerDay); rem %= 365 * minutesPerDay  // this does allow 1y 12mo 4 days, which is acceptable
```
CiaranGruber, line 125: Nice catch here with the edge case. Alternative would be to do 12 * 30 * minutesPerDay but given the way the weeks were handled, this aligns better with the current implementation

> The consideration here is that the next day after 2y is 1y 12mo 4d instead of 1y 11mo 29d - we already have this where the next day for 2mo is currently 1mo 4w 2d instead of 1mo 4w 6d

CiaranGruber (approved):

app/src/main/kotlin/com/fantasyidler/util/Extensions.kt
redjard: yeah. I considered exact gregorian, so 365.2425 * 86400 = 31556952 seconds a year, then divide by 12 for 2629746 seconds (30.436875 days) a month. But that didn't fit the current code. It also would have meant months and years eat partial days, which would seem weird. Only 400y blocks would take an integer number of days.

I doubt anyone who sees multiple years on a timer cares about a few days.
Might be weird if they notice christmas happens on 4 days left then a few years later on 3 days then 2 due to missed leap years.
But then it'd be more noticeable to have a timer line up with their clock, and then go from having 1y 6h at 18:00 to 364d 11h 49m at 18:00 the next day.

PR #1353, opened 2026-08-12
Guide: Fix missing information surrounding prestiging

CiaranGruber:
# Guide: Fixing an out-of-date wiki page surrounding the benefits of prestiging

This pull request shows an example of editing some information in the wiki for an existing page. If you wish to add new pages, this will be included in another guide. Not everything here is perfect, but if you check the reviews in the code, I explain each section of the code and why I did what I did/what the code does.

Additionally, I've gone in a lot of detail in this guide so as to try and cover as much as I could. In reality, especially if you're doing minor changes that just involve adding some sentences or adding strategy guides which don't rely on game data (although more game data is better), then the process should be pretty quick to do and some things like testing although ideal, may not be as necessary.

For steps 2-3, you might find it worthwhile looking the code review I did that explains each section of the code and what it does.

### Step 0: Research, research, research

Before you make any changes, you should ideally make sure you understand the mechanic you're wishing to describe in full. If you have an AI IDE like Cursor, Claude, etc, this can be pretty useful to trawl through the codebase and find the exact piece of code where the mechanic is applied.

If you're not sure how to read the code directly and the mechanic isn't perfectly identifiable, then you could put this as a \#todo item in the generator function, mention it on the page, or potentially put the info you're adding in a tips section at the bottom of a page. If you're writing a strategy guide, then any game data you could include would be great (particularly copying from JSON files).

It is worth noting that because the game receives so many updates, sometimes the wiki or information in the game goes out of date and so checking it in the code is ideal. Ideally, if you can find the actual code that implements what you're discussing, then that's perfect but if you're not sure where it is in the codebase that's alright and we appreciate any contributions you might make (we can always review in the pull request).

### Step 1: Setting up your environment

If you haven't already, then you should set up your Python environment so you can test things out. For small changes like just changing template files (e.g. Adding a strategy guide), this isn't necessary but it's always recommended to make sure you can make sure it all works correctly and see how it looks in the website.

To do this, you should follow the steps in [building/compiling the wiki](https://idlefantasy.tristinbaker.xyz/getting_started_wiki.html#buildingcompiling-the-wiki) which should get you set up and ready to go.

### Step 2: Adding to the template

Once you have the information you need to add to the page, you should find the template associated with the page you wish to edit. In this example, because we're updating the Skills page (in the Prestige section), I've gone and updated the skills page. 

Because I'm also going to be adding dynamically generated data (not the same as dynamic pages) via the agility table, I added a new field (`{agility_table}`) to the template which will be created in the generator function.

[...]

PR #1669, opened 2026-09-02
Add Quest Categories wiki page under Skills

arnoesterhuizen:
## Before submitting

- [x] I have pulled and merged the latest changes from the repository

## Summary

Adds a new "Quest Categories" wiki page under the Skills section, documenting the quest indicator icons (daily, weekly, guild daily, guild progression, main) shown next to skill names in-game.

## Affected pages

- https://idlefantasy.tristinbaker.xyz/Skills (link added to the new page)
- https://idlefantasy.tristinbaker.xyz/QuestCategories (new page)

## Changelog

- Add Quest Categories page documenting the five quest indicator icons and their meanings
- Register the page and its generator in pages.py under the Skills hierarchy
- Link to the Quest Categories page from the Skills page
- Add a Todo note to extract quest categories into data files (the table currently mirrors the QuestCategory enum in CraftingViewModel.kt)

## Anything else?

The quest category names and emoji are hardcoded in the template, mirroring the `QuestCategory` enum in `CraftingViewModel.kt`, since that data is defined in game code rather than a JSON/XML asset. A `# Todo` note has been added in the generator function requesting this be extracted into data files, consistent with similar notes elsewhere in `pages.py`.


wiki/src/pages.py
```diff
@@ -71,6 +71,7 @@ def add_static_pages():
         ]],
         ["Skills", False, [
             ("skills", PageInfo("Skills", "Skills.md", gen_skills)),
+            ("quest_categories", PageInfo("Quest Categories", "QuestCategories.md", gen_quest_categories)),
```
arnoesterhuizen: This registers the new page in the page directory and hierarchy. The tuple is `(page_id, PageInfo(title, filename, generator))`. Placing it here, directly under `skills` and above the `Gathering` subsection, is what makes it appear as a top-level entry within the Skills group in the sidebar. The `page_id` (`quest_categories`) is what you pass to `link()` elsewhere to point at this page.

wiki/src/pages.py
```diff
@@ -820,6 +821,17 @@ def gen_skills() -> str:
     )
 
 
+def gen_quest_categories() -> str:
```
arnoesterhuizen: The generator function. Its name must match the `gen_...` referenced in the PageInfo above. The `# Todo` note follows the convention used elsewhere in this file (e.g. the mercantile shop-bonus and inventory-category notes) for data that is hardcoded in game code and should eventually move to a data file.

wiki/src/pages.py
```diff
@@ -820,6 +821,17 @@ def gen_skills() -> str:
...
+    # Todo: Quest categories (names + emoji) are currently hardcoded in the template, mirroring
+    #  the QuestCategory enum in CraftingViewModel.kt. These should be extracted into XML/JSON
+    #  data files (as with other lists) so the wiki can generate this table from game data.
```
arnoesterhuizen: `get_template(...).format(...)` loads the markdown template and fills in its `{...}` fields. Here the three fields are all page links built with the `link()` helper. `link()` is the correct way to link to other wiki pages — it resolves the page's real URL, so links stay valid even if a filename changes. (For linking to items there is a separate `item_link()` helper.)

arnoesterhuizen (commented):
Adding a few inline notes explaining each part of the change, in the same spirit as the wiki editing guide (#1353), for anyone learning how wiki pages are put together.

wiki/templates/skills/quest_categories.md
```diff
@@ -0,0 +1,27 @@
+# Quest Categories
+
+Quest indicators appear as small superscripted icons next to skill names showing active quest progress. They help players quickly identify which type of quests are active for each skill.
+
+## Quest Category Icons
+
+Quest category icons are defined in `CraftingViewModel.kt` as `QuestCategory`. Each category uses a distinct emoji for visual identification.
+
+| Quest Category | Description |
+|----------------|-------------|
```
arnoesterhuizen: The category table is written as plain markdown rather than being generated, because the categories and emoji are hardcoded in the `QuestCategory` enum in `CraftingViewModel.kt` — there is no JSON/XML asset to read from (yet — see the Todo in the generator). Each row mirrors one enum entry: `DAILY ⏰`, `WEEKLY 📅`, `GUILD_DAILY ⚒️`, `GUILD 🏰`, `MAIN 📜`.

wiki/templates/skills/quest_categories.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Usage in the UI
+
+- **Quest indicators** appear as superscripted counts next to skill names (e.g., ⏰<sup>3</sup> shows 3 active daily quests)
+- Quest indicators are dimmed (38% opacity) when no quests of that category are completable
+- When multiple quests of the same category target the same skill, a superscript shows the count (e.g., 📅<sup>2</sup> for 2 weekly quests)
+
+## See Also
+
+- {skills_link} - Main skills documentation with skill icons
```
arnoesterhuizen: These `{skills_link}`, `{quests_link}` and `{guilds_link}` fields are the placeholders filled in by the generator via `link()`. Everything else on the page is static text that appears as-is. Only pages that actually exist are linked here (Skills, Quests, Guilds); there are no dedicated daily/weekly quest pages to link to.

wiki/templates/skills/skills.md
```diff
@@ -6,6 +6,10 @@ Idle Fantasy has 23 skills split across four categories: gathering, crafting, su
 
 All skills cap at **level 99**.
 
+## Quest Categories
```
arnoesterhuizen, line 9: A short section added to the existing Skills page pointing readers to the new Quest Categories page. This uses a standard markdown link rather than the `link()` helper because it lives in a static template with no matching generator field — the site builder rewrites `QuestCategories.md` to the correct `.html` on build.

wiki/src/pages.py
CiaranGruber: Just wanted to add for anyone reading regarding the # Todo, I think in this case, it would actually be one of the few cases where it is OK staying hardcoded in the wiki. The main reason for this being that in this case, the data does not reflect things that you could change in-game (and it is easy to notice an error if it occurs).

For example, in the gen_equipment() (with the InventoryViewModel), there is the potential for new slots to be added and the shop bonus data could get rebalanced in some way (although this is unlikely)

CiaranGruber:
Nice job, thanks for contributing and also taking the time to set it up so it can be used as an example.

I'll pull this into the repo and can add it to the getting started page as an example pull request. There's a few adjustments I might make before pulling just to align with the other wiki aspects. Normally I'd put these in a review with space for comments, etc, though I mainly want to avoid making this PR messy

Issue #1703, opened 2026-09-05
Remove the time quest dot


Streght:
### Before submitting

- [x] I am on the latest available version (see \#516)
- [x] I searched existing issues and this hasn't been reported

### App Version

1.14.9

### Android Version

17

### Device

Honor Magic 8 Pro

### What happened?

Now that we have icons for the daily and weekly quests, I think the dot for timed quests is no longer relevant and cluttering the interface. 

<img width="1256" height="2808" alt="Image" src="https://github.com/user-attachments/assets/79472639-6b43-4716-8336-6cb9d9510ea4" />

### Steps to reproduce

1. Go to the skills menu and see that each skill with the daily or weekly icon also has the times quest dot

### Expected behavior

Having only the times quest and weekly quest icons

### How often does this happen?

Every time

### Where in the game does this occur?

The skills screen

### Logs or crash trace

```shell

```

CiaranGruber:
Hey, going to close this as a solution is already implemented - if you go to the settings, you can turn off Quest Dots with Skills (See below).

<img width="451" height="490" alt="Image" src="https://github.com/user-attachments/assets/e1ea48ab-bd62-4688-b6cf-a6c114aa0e9d" />

If you don't think this works though, then it'd probably be worthwhile opening up an idea in the #ideas section (or of course GitHub discussion) as to an alternative solution

Streght:
Well hiding this is a workaround, do this dot are only there if there's a quest? If so, it can be removed in my opinion as we have quest icons now, it's only a leftover from before when we didn't have quest markers...
Maybe we can think of a solution rather than closing this right away...

Streght:
@CiaranGruber Maybe you can wait for an answer before closing right away...

CiaranGruber:
Hi @Streght, the standard process (as preferred by Tristin) is for suggestions (i.e. things that might be working as intended but where you might prefer alternatives) to go in Discussions/#idea to bring in other people's ideas which was why I closed it - this setting was also added as a response to the issue you faced (after the icons were introduced) hence why I treated that as the solution

Streght:
@CiaranGruber Sure I understand, maybe I overreacted and I apologize for this.

In my opinion it's not really a bug but it should anyway be removed as it adds confusion and maybe beginners might be wondering what time quest refer to as there's also icons for these now. There's no way to instead remove the icons so you will always have them with the dot and that makes little sense. 

[...]

PR #1715, opened 2026-09-05
New: Remove quest dot feature as superceded by quest indicators

CiaranGruber:
## Before submitting

## Summary

Providews an implementation of a suggestion which removes the quest indicator dot as this has been effectively superceded by the quest indicator icons which shows all daily/guild quests. This relates to to the suggested changes in the referenced issues/discussions. Can be closed if not desired

Relates to closed issue #1703, Closes #1705 (discussion)

## Changelog

- Remove the quest indicator dot and related settings/info
- Add missing guild daily indicators to the slayer task

## Anything else?

The quest indicator dots that were removed:
image

Guild daily icons added:
image

### DIFF
--- a/app/src/main/kotlin/com/fantasyidler/data/model/PlayerModels.kt
+++ b/app/src/main/kotlin/com/fantasyidler/data/model/PlayerModels.kt
@@ -175,8 +175,6 @@ data class PlayerFlags(
     @SerialName("profile_layout") val profileLayout: String = "rail",
     /** Whether session/queue countdowns append the predicted wall-clock completion time. */
     @SerialName("show_session_end_time") val showSessionEndTime: Boolean = true,
-    /** Gold quest dots on the Skills tab icons; some players prefer them off (discussion #1386). */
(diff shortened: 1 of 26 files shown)


nagi-desuuu:
My two cents here is that the existing Quest dots on skills toggle should be adapted to hide quest *indicators* instead, inheriting the "hide on max guild rank/event tokens" behavior, optionally showing everything if the player wants to.

Azerliap:
I agree with @nagi-desuuu here. I will even say, I prefer having the dot and no quest icons at all (in my opinion, they overcrowd the skill a little bit too much, when the skill name is long). Having an option to choose should be preferable.

CiaranGruber:
Thanks for the feedback - there wasn't anything in the referenced discussion so it was good to hear suggestions here. I think perhaps the best way to go would be instead to add a setting which allows independently turning on the indicators and dots separately as it seems some people like indicators and others like the dots.

For now I'll close and probably implement an alternative next week

PR #1738, opened 2026-09-07
Food eating order

WizaHX:
## Before submitting

## Summary

Add an option to chose the order of eating food.

none

## Changelog

- Added option to change the order of eating food, lowest to highest -> Ascending, highest to lowest -> Descending  
- Fixed the order food shows in the combat banner since it was default from inventory(order of obtaining?). Now it follows the selected value/ordering. Same with consumed food, follows same ordering option. Basically everywhere food is shown, is ordered by the selected option. 
- Defaults to the current descending order. 

## Anything else?

I am UI-impaired so i am open to suggestions if the UI is bad and needs change.
Maybe drop-down, maybe toggle/switch, maybe something else :)?
Also if any other valid ordering is suggested? I left alphabetical and unordered(as from inventory) not included since I don't see them as useful.


WizaHX:
changed to dropdown same as arrows and spells, and added 3rd option for least quantity first

PR #1749, opened 2026-09-08
fix and expand wiki content

Michaelomh:
## Before submitting

## Summary

Found some errors in the wiki, decided to update them. Did a whole sweep on the wiki, here is the list of changes.

**Minor Fixes**

- **Farming** - oak ashes yield bonus was rendered as +25%.
- **Mercenary Camp** - contracts last a flat 24h from hire, not "until your next daily reset"
- **Construction** - per-item time is driven by Agility level (60s at Agility 1 → 40s at Agility 99, further reduced by Chronos Spire / Endurance), not a fixed 60s.
- **Spells** - called out that the Void Staff grants infinite runes of *all* types, not just one.
- **Crafting** - dropped "and other items"; every current recipe is jewellery.
- **Slayer shop** - added the missing Slayer Plateskirt and point costs for every shop item.
- **Fletching** - intro now covers magic staves and planks (previously only bows/arrows).

**New content**

- **Prayer** - documented the Church blessings system (XP / Defence / Coin tables, blessing duration and the boosts that extend it, and the Bone Altar combo mechanic.
- **Slayer** - documented the Foretell task-queue system and the Foresight prestige path.
- **Herblore** - explained what an "enhanced" potion actually does and added an *Enhanced Effect* column to the potion table; noted the 1-ash-per-brew cost.
- **Thieving** - noted the stun + skipped frame on a failed pickpocket.
- **Carnival** - noted the Fairgrounds building reduces the active-minigame cooldown (10 → 7.5 → 5 min).

**Contributing docs**

- **Getting Started - Game** - added a "Repository rules" section (localise strings, no DB migrations, no new deps, one fix per PR) and noted an issue is fine for smaller features.
- **Getting Started - Wiki** - added `validation.py` and `wiki_logs.py` to the code-structure list.

Ran `python -m wiki.src validity` - passes.

N/A

## Changelog

- Fixed and expanded wiki content across the Prayer, Slayer, Fletching, Herblore, Thieving, Construction, Crafting, Spells, Mercenary Camp Church, Contributing, Farming, Slayer and Carnival pages

### DIFF
--- a/wiki/src/pages.py
+++ b/wiki/src/pages.py
@@ -898,10 +898,10 @@ def gen_farming() -> str:
     # Ashes tables
     # Todo: Switch to avoid being hardcoded
     ash_rows = []
-    ash_amounts = [("ashes", 1.1), ("oak_ashes", 1.25), ("willow_ashes", 1.35), ("maple_ashes", 1.5), ("yew_ashes", 1.75),
+    ash_amounts = [("ashes", 1.1), ("oak_ashes", 1.2), ("willow_ashes", 1.35), ("maple_ashes", 1.5), ("yew_ashes", 1.75),
                    ("magic_ashes", 2), ("redwood_ashes", 2.5)]
     for ash, bonus in ash_amounts:
-        ash_rows.append([item_name(ash), f"+{int((bonus - 1) * 100)}%"])
+        ash_rows.append([item_name(ash), f"+{round((bonus - 1) * 100)}%"])
 
     magic_bean_note = (
         "Obtaining one requires patience. A lucky harvest may be all it takes. "
@@ -1198,19 +1198,28 @@ def gen_runecrafting() -> str:
 def gen_herblore() -> str:
     recipes = load("recipes/herblore.json")
     assert isinstance(recipes, dict)
+
+    def fmt_effects(effects: dict, enhanced: bool = False) -> str:
+        # Enhanced potions (brewed with an ash catalyst) double every bonus,
+        # floored at base + 1 -- mirrors GameDataRepository.potionEffects.
+        def val_for(v):
+            return max(int(v * 2), v + 1) if enhanced else v
+        return ", ".join(f"{stat.title()} +{val_for(val)}" for stat, val in effects.items())
+
     rows = sorted(
         [[
             item_name(k),
             r["level_required"],
             fmt_materials(r["materials"]),
-            ", ".join(f"{stat.title()} +{val}" for stat, val in r.get("effects", {}).items()),
+            fmt_effects(r.get("effects", {})),
+            fmt_effects(r.get("effects", {}), enhanced=True),
             r["xp_per_item"],
         ] for k, r in recipes.items()],
         key=lambda r: r[1]
     )
     return get_template("skills/crafting/herblore").format(
         icon=html_image(skill_icon_path("herblore"), "", "text"),
-        potion_table=table(['Potion','Level','Ingredients','Effect','XP'], rows),
+        potion_table=table(['Potion','Level','Ingredients','Effect','Enhanced Effect','XP'], rows),
     )
 
 
@@ -1259,8 +1268,40 @@ def gen_thieving() -> str:
     )
 
 
+# Todo: Switch to avoid being hardcoded (mirrors ChurchRepository.ALL_BLESSINGS
+# and ChurchRepository.boneCostFor, which live in Kotlin, not game data).
+_BLESSING_BONE_COST = {1: 10, 10: 20, 20: 35, 30: 55, 40: 80, 50: 110,
+                       60: 145, 70: 185, 80: 230, 90: 265, 99: 300}
+_BLESSINGS = {
+    "XP": [  # (name, prayer level, XP multiplier)
+        ("Blessed Focus", 1, 1.05), ("Blessed Focus II", 10, 1.10),
+        ("Blessed Focus III", 20, 1.15), ("Tithe Blessing", 30, 1.18),
+        ("Tithe Blessing II", 40, 1.20), ("Tithe Blessing III", 50, 1.25),
+        ("Divine Focus", 60, 1.28), ("Divine Focus II", 70, 1.32),
+        ("Divine Grace", 80, 1.37), ("Divine Grace II", 90, 1.43),
+        ("Sacred Grace", 99, 1.50),
+    ],
+    "DEFENSE": [  # (name, prayer level, flat defence)
+        ("Stone Skin", 1, 2), ("Stone Skin II", 10, 4), ("Stone Skin III", 20, 6),
+        ("Stone Skin IV", 30, 9), ("Iron Ward", 40, 12), ("Iron Ward II", 50, 15),
+        ("Diamond Skin", 60, 18), ("Diamond Skin II", 70, 22),
+        ("Holy Shield", 80, 26), ("Holy Shield II", 90, 30), ("Aegis", 99, 35),
+    ],
+    "COINS": [  # (name, prayer level, coin bonus fraction)
+        ("Fortune I", 30, 0.08), ("Fortune II", 40, 0.10), ("Fortune III", 50, 0.13),
+        ("Fortune IV", 60, 0.15), ("Fortune V", 70, 0.18), ("Abundance", 80, 0.20),
+        ("Abundance II", 90, 0.23), ("Abundance III", 99, 0.25),
+    ],
+}
+
+
+def _blessing_table(kind: str, effect_fmt) -> str:
+    rows = [[name, level, _BLESSING_BONE_COST[level], effect_fmt(mag)]
+            for name, level, mag in _BLESSINGS[kind]]
+    return table(["Name", "Prayer Level", "Cost (bones)", "Effect"], rows)
+
+
 def gen_prayer() -> str:
-    # Todo: Add info about bone altar
     bones = load("bones.json")
     assert isinstance(bones, dict)
     rows = sorted(
@@ -1271,6 +1312,9 @@ def gen_prayer() -> str:
     return get_template("skills/support/prayer").format(
         icon=html_image(skill_icon_path("prayer"), "", "text"),
         prayer_table=table(['Bone / Ash','XP Each'], rows),
+        blessing_xp_table=_blessing_table("XP", lambda m: f"+{round((m - 1) * 100)}% XP"),
+        blessing_defense_table=_blessing_table("DEFENSE", lambda m: f"+{m} Defence"),
+        blessing_coins_table=_blessing_table("COINS", lambda m: f"+{round(m * 100)}% coins"),
     )
 
 
--- a/wiki/src/pages.py
+++ b/wiki/src/pages.py
@@ -21,7 +21,7 @@ from wiki.src.game_data import STRINGS, load, title, item_name, house_item_name,
     trade_route_name, thieving_npc_name, quest_name, agility_course_name, town_building_name, quest_desc, title_name, \
     pet_name, boss_name, boss_desc, trade_route_desc, pet_desc, item_desc, dungeon_name, dungeon_desc, expedition_name, \
     expedition_desc, seasonal_event_name, seasonal_reward_desc, prestige_effect_desc, tree_name, merc_name, race_name, \
-    carnival_prize_name, carnival_prize_desc, slot_name
+    carnival_prize_name, carnival_prize_desc, slot_name, blessing_name
 from wiki.src.page_hierarchy import PageHierarchy
 from wiki.src.wiki_logs import LOGGER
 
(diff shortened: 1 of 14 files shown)


wiki/templates/contributing/getting_started_game.md
```diff
@@ -28,14 +28,23 @@ Developing new features is a fun way to add to the game and see the changes you'
 
 If you want to develop a new feature you should follow the below steps:
 
-1. Create a new discussion topic in [discussions](https://github.com/tristinbaker/IdleFantasy/discussions). This ensures that the community is able to chip in and provide feedback before you start development. If this is a significant new feature, you could consider adding some example designs of what you're thinking to help provide more clarity.
+1. Create a new discussion topic in [discussions](https://github.com/tristinbaker/IdleFantasy/discussions). This ensures that the community is able to chip in and provide feedback before you start development. If this is a significant new feature, you could consider adding some example designs of what you're thinking to help provide more clarity. For a smaller feature, opening a GitHub issue (as the repository's `CONTRIBUTING.md` describes) is also fine.
```
CiaranGruber: I believe @tristinbaker wants features to be solely in discussions/#ideas, if the CONTRIBUTING.md suggests to open issues, I think that might be worth double checking and then fixing in the CONTRIBUTING.md file if needed

wiki/templates/contributing/getting_started_game.md
```diff
@@ -28,14 +28,23 @@ Developing new features is a fun way to add to the game and see the changes you'
...
 4. Then, once all the merge conflicts have been sorted, you can test to ensure the feature still works properly.
 5. Finally, once you've tested and synced your fork, you can create a pull request, explaining what you've added and provide a reference to the relevant discussion.
 
 It's quite common while developing an idea that you come up with lots of additional changes and revisions. If these are large, you should consider updating the discussion topic so it properly reflects the idea in full.
 
+## Repository rules
+
+A few hard rules from the repository's [`CONTRIBUTING.md`](https://github.com/tristinbaker/IdleFantasy/blob/main/CONTRIBUTING.md) apply to every change, and a pull request that breaks one will need reworking:
+
+- **Localise user-visible strings.** If you add or change any string a player can read, propagate it to every locale file under `app/src/main/res/values-*/strings.xml`, not just the base `values/strings.xml`.
```
CiaranGruber: I think this should probably mention the `normalise_locale_strings.py` script which does this automatically for people which I would think is preferable over potentially missing things when doing it manually or with AI as this wasn't initially a thing when the `CONTRIBUTING.md` was made - the `CONTRIBUTING.md` could probably be updated as well for this.

Probably worth having @tristinbaker confirm this first though as he might prefer things being loaded as part of the `release.sh` script

wiki/templates/skills/support/prayer.md
```diff
@@ -5,3 +5,37 @@ Bury bones to earn Prayer XP. Higher-tier bones give more XP. Ashes from Firemak
 {prayer_table}
 
 > Tip: Dragon and giant bones give the most XP per session. Farm higher-level dungeons for better bone drops.
+
+## Blessings
+
+Blessings are temporary buffs activated at the **Church** building and paid for in bones. Only one blessing can be active at a time; activating another replaces it. The bone cost is fixed per tier and is paid in regular-bone equivalents (higher-tier bones count for more, and your most valuable bones are spent first).
```
CiaranGruber: Potentially link to the church building (I think `link("buildings", "Church", "church")` should do this in the generator function)?

wiki/templates/skills/support/prayer.md
```diff
@@ -5,3 +5,37 @@ Bury bones to earn Prayer XP. Higher-tier bones give more XP. Ashes from Firemak
...
+{blessing_defense_table}
+
+### Coin blessings - bonus coins earned
+
+{blessing_coins_table}
+
+### Boosting blessings
+
+- **Wear the Prayer cape** - increases blessing strength (roughly 1.5× the listed effect).
+- **Upgrade the Church** - each tier adds to the base 24-hour duration: +6h at tier 1, +12h at tier 2, +24h at tier 3.
```
CiaranGruber: Ideally could link to the buildings page so people can see the church table but should be fine otherwise

wiki/templates/skills/support/prayer.md
```diff
@@ -5,3 +5,37 @@ Bury bones to earn Prayer XP. Higher-tier bones give more XP. Ashes from Firemak
...
+
+{blessing_defense_table}
+
+### Coin blessings - bonus coins earned
+
+{blessing_coins_table}
+
+### Boosting blessings
+
+- **Wear the Prayer cape** - increases blessing strength (roughly 1.5× the listed effect).
```
CiaranGruber: I don't think this value is accurate (or potentially could have a typo). Tbh I am a bit confused as the UI shows it as 1.15x mult but it appears to be more like 5.3% - it's probably worthwhile checking in the code to confirm

wiki/templates/town/carnival.md
```diff
@@ -8,7 +8,7 @@ The Carnival is a Town minigame hub where you earn Carnival Tickets and spend th
 
 {idle_table}
 
-**Active minigames** are played directly and reward more tickets on Hard difficulty. Each has its own 10-minute cooldown between attempts, so you can rotate between all of them back-to-back.
+**Active minigames** are played directly and reward more tickets on Hard difficulty. Each has its own cooldown between attempts — 10 minutes by default — so you can rotate between all of them back-to-back. Upgrading the Fairgrounds building shortens this: 7.5 minutes at tier 1, down to 5 minutes at tier 3.
```
CiaranGruber: Perhaps as a more future-proofed approach (in case new tiers, etc, get added in the future), it could be worth changing this to the below suggestion and then adding a `link("buildings", "Upgrading the fairgrounds", "fairgrounds")` reference in the generator function.
```suggestion
**Active minigames** are played directly and reward more tickets on Hard difficulty. Each has its own cooldown between attempts — 10 minutes by default — so you can rotate between all of them back-to-back. {fairground_upgrade_page} building provides bonuses such as additional games and reduced cooldowns.
```

wiki/templates/combat/spells.md
```diff
@@ -4,6 +4,6 @@ Equip a magic weapon and select a spell before entering a dungeon or boss fight.
 
 {spell_table}
 
-> Some staves provide **infinite runes** of a specific type, removing the rune cost for that spell.
+> Some staves provide **infinite runes** of a specific type, removing that rune's cost from any spell that uses it. The Void Staff provides **infinite runes** for all types.
```
CiaranGruber: Probably worth linking to the void staff entry in the equipment page which you can do by applying the below change and adding `item_link("void_staff")` in the generator function
```suggestion
> Some staves provide **infinite runes** of a specific type, removing that rune's cost from any spell that uses it. The {void_staff_link} provides **infinite runes** for all types.
```

wiki/src/pages.py
```diff
@@ -1259,8 +1266,38 @@ def gen_thieving() -> str:
     )
 
 
+_BLESSING_BONE_COST = {1: 10, 10: 20, 20: 35, 30: 55, 40: 80, 50: 110,
+                       60: 145, 70: 185, 80: 230, 90: 265, 99: 300}
+_BLESSINGS = {
```
CiaranGruber, line 1273: Worthwhile keeping the comment as this just adds a reminder to fix in the future - I'll probably fix this issue sometime soon but for now, good to have the Todo as a reminder

```suggestion
# Todo: Move blessings from ChurchRepository.ALL_BLESSINGS into their own dedicated blessings.json file and reference in the game with gameData and here by loading the json file
_BLESSINGS = {
```

wiki/src/pages.py
```diff
@@ -1198,19 +1198,26 @@ def gen_runecrafting() -> str:
...
+            fmt_effects(r.get("effects", {})),
+            fmt_effects(r.get("effects", {}), enhanced=True),
             r["xp_per_item"],
         ] for k, r in recipes.items()],
         key=lambda r: r[1]
     )
     return get_template("skills/crafting/herblore").format(
         icon=html_image(skill_icon_path("herblore"), "", "text"),
-        potion_table=table(['Potion','Level','Ingredients','Effect','XP'], rows),
+        potion_table=table(['Potion','Level','Ingredients','Effect','Enhanced Effect','XP'], rows),
```
CiaranGruber, line 1220: Just flagging for @tristinbaker to review and consider - personally I feel the extra column looks a little unnecessary/messy since the wiki page already mentions that enhancing potions applies a 2x mult though that's just my personal thoughts.

See page below:

Image

No need to change for now @Michaelomh unless you think it'd also be better without

wiki/src/pages.py
```diff
@@ -1259,8 +1266,38 @@ def gen_thieving() -> str:
...
+        ("Stone Skin IV", 30, 9), ("Iron Ward", 40, 12), ("Iron Ward II", 50, 15),
+        ("Diamond Skin", 60, 18), ("Diamond Skin II", 70, 22),
+        ("Holy Shield", 80, 26), ("Holy Shield II", 90, 30), ("Aegis", 99, 35),
+    ],
+    "COINS": [
+        ("Fortune I", 30, 0.08), ("Fortune II", 40, 0.10), ("Fortune III", 50, 0.13),
+        ("Fortune IV", 60, 0.15), ("Fortune V", 70, 0.18), ("Abundance", 80, 0.20),
+        ("Abundance II", 90, 0.23), ("Abundance III", 99, 0.25),
+    ],
+}
```
CiaranGruber, line 1293: Ideally, rather than using the full English names, the best approach is to use their IDs (like it is in the `ChurchRepository.ALL_BLESSINGS` and then create a new string loader in `game_data.py` like it's done [here](https://github.com/tristinbaker/IdleFantasy/blob/3f3de3f11e2ca82450381ec48bf7f3a2182019e7/wiki/src/game_data.py#L390-L391). This will need to refer to the strings [here](https://github.com/tristinbaker/IdleFantasy/blob/3f3de3f11e2ca82450381ec48bf7f3a2182019e7/app/src/main/res/values/strings.xml#L1058-L1087) which the standard string resolution does.

You should then be able to refer to this function like blessing_name(blessing_id: string) -> string

CiaranGruber (changes requested):
Hey thanks for the updates, I think a lot of them are definitely necessary as there's plenty in the wiki that needs working on so appreciate the time.

My only concern not in the comments is that having exact figures in the markdown files (that aren't autogenerated) could potentially lead to drift between the game and the wiki should these figures ever change in the future although there's definitely a balance and having some is good for clarity and I don't think you need to change those parts unless you deem it necessary.

I've put a few comments to try and improve the pull request by using the game data, strings, etc. Since I assume you're generally familiar with things (I know you've contributed plenty before), I haven't given too much description for some comments though if anything seems unclear, feel free to reply and I can clarify

wiki/templates/skills/support/prayer.md
Michaelomh: updated!

wiki/templates/skills/support/prayer.md
Michaelomh: updated!

wiki/templates/skills/support/prayer.md
Michaelomh: updated! I just remove the listed effect and just say that it increases blessing strength.

wiki/templates/town/carnival.md
Michaelomh: updated!

wiki/templates/combat/spells.md
Michaelomh: updated!

wiki/src/pages.py
Michaelomh: updated!

wiki/src/pages.py
Michaelomh: I'll keep it. It mentioned really briefly, literally - "Brewing a potion with any ash as catalyst (ash tier has no difference) can produce an enhanced potion with **double the effects.**". My explanation is way clearer. 

Also, it removes any doubt that it might have different effects for stronger potions (sometimes this is the case) Like overload potion instead of +20, would be +25 cause of "end game".

CiaranGruber: Oh yeah, I agree the explanation is clearer, my comment was just about the column specifically but yeah, should be all good

wiki/templates/contributing/getting_started_game.md
Michaelomh: Let me remove this if it makes it easier.

wiki/templates/contributing/getting_started_game.md
Michaelomh: I'll remove this.

wiki/src/pages.py
Michaelomh: updated!

wiki/templates/contributing/getting_started_game.md
CiaranGruber: Ah yeah, I mean I thought the general section was good though all good either way. Even so, it did point out an issue with the CONTRIBUTING.md having potentially outdated info so that could do with updating - that can maybe be done in a future PR though so we can get this one going though

wiki/src/pages.py
```diff
@@ -2033,6 +2077,7 @@ def gen_carnival() -> str:
     ]
 
     return get_template("town/carnival").format(
+        fairgrounds_link=link("buildings", "Fairgrounds"),
```
CiaranGruber: Small addition to apply before it gets pulled in
```suggestion
        fairgrounds_link=link("buildings", "Fairgrounds", "fairgrounds"),
```

wiki/src/pages.py
CiaranGruber:
```suggestion
        [f"Higher or Lower ({link("buildings", "Fairgrounds", "fairgrounds")} tier 2+)", "Guess higher or lower over several rounds — more correct in a row pays more", "up to 5", "up to 8"],
```

CiaranGruber:
```suggestion
        [f"Pick-a-Cup ({link("buildings", "Fairgrounds", "fairgrounds")} tier 1+)", "Track which cup hides the gem through a shuffle", "4", "7"],
```

CiaranGruber (approved):
Looks good, appreciate the time to fix up the pages and also checkout out the review afterwards. I've got a couple things listed but I'll simply apply these as a batch commit before pulling the changes in

PR #1754, opened 2026-09-09
feature: show (Max) on fully maxed prestige skills

Michaelomh:
## Before submitting

## Summary

Displays "⭐x5 (Max)" when a skill has reached max prestige level.

❗Take note the number of stars might not be accurate. I use the same conditional as the check to display the text "Maximum prestige reached. Every point this skill can earn is yours." (maximum number of lifetime points). And the save file is generated. 
image

## Changelog

- Skills and Combat screens show "(Max)" beside the prestige stars when a skill's prestige is fully maxed


CiaranGruber (approved):
Thanks for contributing and looks like it should be good from a code standpoint to merge.

Some potential considerations:
- Perhaps should the prestige button change it's string to "Skill tree" considering you can't prestige any more? - This does have the potential for making the UI look a little more cluttered however

Checks done:
- Visual review of changes
- Run the code on Android Studio and tested that the feature works as expected

PR #1760, opened 2026-09-10
Add a wiki guide with a JS utility to estimate odds in the infinite tower

adnm9:
## Before submitting

- [x ] I've tested any changes with the appropriate tests (eg. Unit tests, validation scripts)
- [x ] I've pulled and merged the latest changes from the repository

## Summary

This PR adds a new wiki page which contains a JS simulator to estimate odds in the infinite tower given the player levels, gear, and food.

Relates to #1744 

## Changelog

- Adds a player guide page to the wiki which contains a utility to allow players to simulate tower fights at different loadouts.

## Anything else?

This wiki page builds successfully and works. However, when running the validity script, the following line appears:

CRITICAL:root:The page `guide_tower_odds_sim` references a non-existent page: guide_the_infinite_tower.html.md

This is a bogus warning, as the reference is to guide_the_infinite_tower.html,
https://github.com/adnm9/IdleFantasy/blob/d88a02a87bf5288a3faaa2d545ab538a4d30f66e/wiki/templates/guides/tower_odds_sim.md?plain=1#L19
which exists and loads just fine. I don't know where the .md is getting added in the generation process. It could be that it is just warning because the relative link doesn't go anywhere in the md, and only works in the built website.

### DIFF
--- a/wiki/src/pages.py
+++ b/wiki/src/pages.py
@@ -2303,6 +2303,18 @@ def gen_guide_tower_odds_sim(guide: str) -> str:
     enemies_data = load("enemies.json")         # Bring in enemy stats
     food_map_placeholder = "{FOODMAP}"          # Placeholder in md file
     enemies_placeholder  = "{ENEMIES}"          # Placeholder in md file
+
+    # -----------------------------------------------------------------
+    # OPTIMIZATION: Define and use the list of valid tower enemies,
+    # reducing website bandwidth
+    TOWER_ENEMIES = {
+        "goblin", "skeleton", "zombie",
+        "orc_warrior", "dark_wizard", "bandit",
+        "cave_troll", "shadow_beast", "demon",
+        "forge_demon", "shadow_assassin", "abyssal_leech",
+        "void_stalker", "void_guardian", "abyssal_lord",
+        "void_archon", "eternal_sentinel"
+    }
     
     # Create the food map in the same format as the original JS
     food_map_data = {}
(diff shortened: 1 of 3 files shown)


adnm9:
Update: The validity error has been eliminated.

wiki/src/pages.py
```diff
@@ -2284,8 +2284,102 @@ def gen_housing() -> str:
...
+    guide = before_placeholder + food_map + after_placeholder
+    
+    # Replace the enemies data placeholder with the enemies data
+    parts = guide.split(enemies_placeholder, 1)
+    before_placeholder  = parts[0]
+    after_placeholder   = parts[1]    
+    guide = before_placeholder + enemies_string + after_placeholder
+
+    # Can simply return the updated guide string, the rest of the formatting is handled by other functions        
+    return guide
```
CiaranGruber: Rather than this approach, I'd recommend using Python's integrated .format() function as this is a neat method to fill in placeholders and is why the wiki uses fields in that format. To do this, you should be able to do this instead (although make sure to check).

Additionally, there's a number of helper functions that are used in the wiki (in the `pages.py` file) which can do a number of things like linking to other pages without causing the errors you had below and also working for both the GitHub wiki and website (which use different types of links).
```suggestion
    # Can simply return the updated guide string, the rest of the formatting is handled by other functions        
    return guide.format(
        food_map=food_map,
        enemies=enemies,
        infinite_tower_guide=link("guide_the_infinite_tower")
    )
```

CiaranGruber:
Hey @adnm9, thanks for taking the time to put this together, I think the idea should be good though you have actually shown a missing feature in the wiki being that HTML pages aren't currently supported properly (while your method does work, there's definitely ways we can expand the wiki to make it fit a bit better).

Prior to pulling this in there'll be a few features to add to the wiki to make sure that things can be pulled in smoothly. For now, we can leave the pull request open as-is unless Tristin prefers closing though even if it does get closed, we'll definitely look to integrate this page into the wiki though it may take a little bit for things to be ready

adnm9:
Sounds good, thanks for the update.

PR #1792, opened 2026-09-13
Refactor: Move church blessings into blessings.json

Michaelomh:
## Before submitting

## Summary

- Refactor Church blessings to into assets/data `blessings.json`
- No functional changes. No UI changes.

**Wiki still works**
image

Closes https://github.com/tristinbaker/IdleFantasy/pull/1795 as this includes those commits

## Changelog

- Updated all church blessing from hardcoded list to use the new json data instead


app/src/main/kotlin/com/fantasyidler/repository/ChurchRepository.kt
```diff
@@ -24,65 +24,31 @@ class ChurchRepository @Inject constructor(
     private val townRepoProvider: Provider<TownRepository>,
     private val buffNotifScheduler: BuffNotificationScheduler,
     private val boostRepo: BoostRepository,
+    private val gameData: GameDataRepository,
 ) {
     /** Bone cost after the gnome Trickster's Favor prestige discount. */
     fun discountedBoneCost(blessing: BlessingData, flags: PlayerFlags): Int =
         discountedBoneCost(blessing, boostRepo.blessingCostMultiplier(flags))
 
     companion object {
```
CiaranGruber: Since it no longer needs the ALL_BLESSINGS value, I think perhaps the best way is to have all functions and have the churchRepo injected instead  (like how other repositories are injected) as this follows the approach in the rest of the codebase

app/src/main/kotlin/com/fantasyidler/repository/ChurchRepository.kt
```diff
@@ -24,65 +24,31 @@ class ChurchRepository @Inject constructor(
...
-            BlessingData("abundance_ii",       90, BlessingType.COINS,   0.23f),
-            BlessingData("sacred_grace",       99, BlessingType.XP,      1.50f),
-            BlessingData("aegis",              99, BlessingType.DEFENSE, 35f),
-            BlessingData("abundance_iii",      99, BlessingType.COINS,   0.25f),
-        )
-
-        private val BY_KEY = ALL_BLESSINGS.associateBy { it.key }
-
-        fun activeBlessing(flags: PlayerFlags): BlessingData? {
+        fun activeBlessing(flags: PlayerFlags, blessings: List<BlessingData>): BlessingData? {
```
CiaranGruber, line 34: Relating to what I refered to earlier, I would put these as normal functions, not part of the companion object as you then don't need to pass in the blessings every time (it can just read gameData.blessings), though this isn't my area of expertise so curious whether you had a particular reason to keep it like that or was it just to minimise changes?

If not, you can maybe leave as is and I'll make a separate pull request with that being the main difference as we can then leave it up to Tristin who probably knows more about it.

Edit: Just realised I wasn't too clear about the functions I meant, I meant specifically the ones before `effectiveMagnitude` as these rely on gameData

CiaranGruber (commented):
Nice job, got there before me. There is just one thing I'm a little unsure about which is whether to keep things in the companion object. Would be good to hear your thoughts if you had a particular reason. Only suggestion to update is that it could be good to also update the wiki to read the json file but it doesn't change anything so can fix if you wanted or it can be done in a future PR

app/src/main/kotlin/com/fantasyidler/repository/ChurchRepository.kt
Michaelomh: I didn't really think about it much when refactoring the code. One thing that I don't really like to do is refactor something big like this without the proper unit tests in place since it's something that i'm not familiar with (and I might break something). I'll leave it as-is for now.

CiaranGruber: Yeah nah, I think that's the best approach. There were some cyclic dependency issues that needed lazy injection so can just let Tristin decide what he prefers. Might be worth adding "Closes #1795" like I've done to the Related issues/discussions which was the pull request I mentioned so it does one or the other

CiaranGruber (approved):
Looks good, I ended up closing the other PR as it caused a lot of issues with cyclic dependencies and test case issues so I think this approach is cleaner/more minimal

PR #1814, opened 2026-09-16
Remove misleading XP boost message from session summary page

Michaelomh:
## Before submitting

## Summary

- Removes the "2× All skills XP was active" line from the session summary and worker summary dialogs.

image

## Changelog

- Remove misleading "2× All skills XP was active" text from session/worker summary dialogs

## Anything else?


CiaranGruber:
From looking at the code, it seems like the feature is working as intended - it's just that the string is misleading as it checks for **any** 2x boost being active but the string implies it's **just** the All Skills boost (as a remnant from when this was the only boost present). Perhaps a better change that would clarify the string without changing the functionality would be "2x XP boosts were active" (or you could change it to "Timed boosts were active" to be more generic though this should also include the blessing mult then).

An alternative solution could also be to restrict it be only the global 2x boost rather than any boost being active although given the code was specifically changed by Tristin to include the prestige boosts, I think it was just that fixing the strings was missed

PR #1846, opened 2026-09-20
bugfix: fix misleading XP boost message on session summary

Michaelomh:
## Before submitting

## Summary

- Rather than removing the line (as proposed in #1814), we would be reworking the text to be more accurate in the summary page.

Relates to #1814


CiaranGruber (changes requested):
Looks good though one thing to update - whenever the English strings are modified, the other languages should probably be marked as well for changes as otherwise these could slip through the cracks. My typical approach is to change them back to the English variant with the untranslated marker so they appear correct in-game and are obvious translation targets for people - an alternative is just to mark as untranslated so they show in the translating website but I think this has its drawbacks

